PR #22, opened 2022-12-02
Update cell size and define laser field at the cell centers

RemiLehe:
With the current definition of `box.dx` (which is based on `np.linspace`), the box cell size is actually not `(hi-lo)/npoints` but `(hi-lo)/(npoints-1)`. This might confuse the user. 

This corrects the above by defining `box.dx` to be explicitly `(hi-lo)/npoints`. In addition, it defines the fields at the cell-centers, which might be more intuitive to the user. (Defining the fields at the nodes would require `npoints+1`.)

### DIFF
--- a/lasy/utils/box.py
+++ b/lasy/utils/box.py
@@ -37,5 +37,6 @@ class Box:
         self.axes = []
         self.dx = []
         for i in range(self.ndims):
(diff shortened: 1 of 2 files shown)


MaxThevenet:
To be discussed and fixed

Issue #23, opened 2022-12-05
Add example how to read and plot lasy data with e.g. openPMD


MaxThevenet:
For convenience, we could store this somewhere or append it to `test.py`.
Also, should we add capabilities for openPMD-viewer to work with envelope data (e.g. in LpaDiagnostics)?

PR #46, opened 2022-12-12
Added openpmd Examples to test.py  #23

KPeterBlum:
I added some plots with openPMD-viewer to the test.py file. It currently just saves the plots in the same folder.
It is one plot for both real and imaginary part for each of the geometries.
I guess that's all there is to plot at the moment.


MaxThevenet:
Feature tested separately

PR #78, opened 2023-01-23
Add super-Gaussian transverse laser-profile

lucafedeli88:
This Work-In-Progress PR adds  super-Gaussian transverse laser-profiles.
I think that there are a couple of points to discuss:
- ~~is it OK to call the shape parameter of the super-gaussian "β" ? It is what scipy does for the generalized normal distribution ( https://docs.scipy.org/doc/scipy/tutorial/stats/continuous_gennorm.html )~~ we agreed to call this parameter `n_order`
- ~~the idea is to always have the maximum of the function equal to one, right (i.e., we do not normalize the functions)~~ ? 
- ~~should we check for the validity of the profile parameters (e.g., if w_0 > 0)~~ ? we do not check the validity of the parameters


MaxThevenet (approved):
Looks good to me, thanks for this PR!

lucafedeli88:
sorry, something weird has happened ! I will re-open the PR...

PR #97, opened 2023-02-10
Add transverse laser from data and offset profile functionality

rob-shalloo:
simple class to import a transverse laser profile from data. It takes a 2D intensity image together with the lo and hi ends of the physical domain on which the data is defined. Importantly, the data must already be imported into a numpy array and pre-processed to have the background removed. This should be done outside this class and we can provide some functionality for it in utils. 

Another optional argument is "center_beam" which is true by default and centers the beam beam based on the center of mass of the data. This is useful when pulling in images from cameras where the beam position is arbitrary, but may want to be set to False when the laser beam position in the data is important. 

The profiles are normalised such that the 2D integral of the intensity is 1.

### DIFF
--- a/lasy/profiles/transverse/transverse_profile.py
+++ b/lasy/profiles/transverse/transverse_profile.py
@@ -12,10 +12,14 @@ class TransverseProfile(object):
     def __init__(self):
         """
         Initialize the transverse profile.
-        """
-        pass
 
-    def evaluate(self, x, y):
+        Here we initialise x and y spatial offsets as placeholders
+        """
+        self.x_offset = 0
+        self.y_offset = 0
+        
+
+    def _evaluate(self, x, y):
         """
         Returns the transverse envelope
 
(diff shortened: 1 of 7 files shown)


rob-shalloo:
After discussions during the hackathon we added some general functionality to the laser transverse profile class to be able to spatially offset any transverse profile in a general way.

lasy/profiles/transverse/transverse_profile.py
```diff
@@ -34,3 +37,44 @@ def evaluate(self, x, y):
         # The base class only defines dummy fields
         # (This should be replaced by any class that inherits from this one.)
         return np.zeros(x.shape, dtype="complex128")
+
+    def evaluate(self, x, y):
+        """
+        Returns the transverse envelope modified by any spatial offsets.
+        This is the public facing evaluate method.
```
MaxThevenet:
```suggestion
        This is the public facing evaluate method, it calls the _evaluate function of the derived class.
```

lasy/profiles/transverse/transverse_profile.py
```diff
@@ -34,3 +37,44 @@ def evaluate(self, x, y):
...
+        Returns
+        -------
+        envelope: ndarray of complex numbers
+            Contains the value of the envelope at the specified points
+            This array has the same shape as the arrays x, y
+        """
+
+        return self._evaluate(x + self.x_offset, y + self.y_offset)
+
+    def offset(self, x_offset, y_offset):
```
MaxThevenet:
```suggestion
    def set_offset(self, x_offset, y_offset):
```

lasy/utils/exp_data_utils.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+
+
+def find_com(img):
```
MaxThevenet: Maybe this? We probably don't call it so often that we need a short name
```suggestion
def find_center_of_mass(img):
```

MaxThevenet (approved):

PR #104, opened 2023-02-10
add test for transverse profiles

MaxThevenet:
This PR proposes to test all transverse profiles, and then remove duplicate tests.

Points below are potential improvements, but they will be kept for a separate PR.
- [ ] `test_profile_gaussian_3d_cartesian` and `test_profile_gaussian_cylindrical` can be kept, as they do full 3D profiles, but we do not have to write to file and propagate, but instead check width and duration vs. theory. That tests Gaussian profile
- [ ] Add a test for `CombinedLongitudinalTransverseProfile`, likewise checking vs. theory (so probably from a simple profile). That tests this combined profile.

Separately a propagator test has been updated. Once the points + few other things are implemented, the coverage should be better.


MaxThevenet:
This PR should be ready for review.

tests/test_laser_profiles.py
```diff
@@ -41,6 +46,92 @@ def gaussian():
...
+    std = np.sqrt(np.average(r**2, weights=np.abs(field)))
+    print("std_th = ", std_th)
+    print("std = ", std)
+    assert np.abs(std - std_th) / std_th < 0.01
+
+    # LaguerreGaussianTransverseProfile
+    print("LaguerreGaussianTransverseProfile")
+    p = 2
+    m = 0
+    std_th = np.sqrt(5 / 2) * w0
```
RemiLehe, line 68: I added the exact prediction here.

tests/test_laser_profiles.py
```diff
@@ -41,6 +46,92 @@ def gaussian():
...
+    profile = LaguerreGaussianTransverseProfile(w0, p, m)
+    field = profile.evaluate(r, np.zeros_like(r))
+    std = np.sqrt(np.average(r**2, weights=r * np.abs(field) ** 2))
```
RemiLehe, line 71: I modified this formula (`np.abs(field) -> r*np.abs(field)**2`) so that it has an easy analytical solution.

tests/test_laser_profiles.py
```diff
@@ -41,6 +46,92 @@ def gaussian():
...
+
+def test_transverse_profiles_3d():
+    npoints = 200
+    w0 = 10.0e-6
+
+    # HermiteGaussianTransverseProfile
+    print("HermiteGaussianTransverseProfile")
+    n_x = 2
+    n_y = 2
+    std_th = np.sqrt(5.0 / 4) * w0
```
RemiLehe, line 106: I added the theoretical expression here.

tests/test_laser_profiles.py
```diff
@@ -41,6 +46,92 @@ def gaussian():
...
+    profile = HermiteGaussianTransverseProfile(w0, n_x, n_y)
+    x = np.linspace(-4 * w0, 4 * w0, npoints)
+    y = np.zeros_like(x)
+    field = profile.evaluate(x, y)
+    std = np.sqrt(np.average(x**2, weights=np.abs(field) ** 2))
```
RemiLehe, line 111: I modified this (`np.abs(field) -> np.abs(field)**2`) so that it has an easy analytical solution

RemiLehe (approved):

PR #108, opened 2023-02-10
Improve documentation

AngelFP:
This PR introduces a few changes to the documentation:
* `napoleon` has been replaced by `numpydoc`. See features here (https://numpydoc.readthedocs.io/en/latest/format.html)
* Add `matplotlib.sphinxext.plot_directive` in order to be able to include plots defined by code in the documentation.
* Set copyright year automatically.
* Add `Laser` to API reference.
* Add examples to `Laser` and `GaussianProfile`.
* Fixed many formatting issues.
* Moves the definition of the class parameters to the class docstring, instead of the `__init__` docstring, as recommended by `numpydoc`. This allows us to actually include examples in the class documentation that appear after the parameters.


AngelFP:
There are two issues from CodeQL, but they are not related to changes I've made (I think). Not sure where they are coming from.

AngelFP:
Hmm, it seems my PR somehow reintroduced some code that had been deleted previously today. That seems to have been the issue. It should now be fine after the last commit.

RemiLehe:
Thanks for fixing this! Sorry for the messy process, when I merged `hackathon2`!

RemiLehe (approved):

PR #110, opened 2023-02-10
Add addition and scaling operators to base Profile class

soerenjalas:
This PR defines the `__add__`, `__mul__` and `__rmul__` of the base Profile class. 

For this two new subclasses are defined:

* `SummedProfile` which takes a list of Profiles and provides an evaluate function that returns the sum of the individual evaluations.
* `ScaledProfile` which takes a Profile and a scalar factor. The evaluate function then returns the product of the original profile and the factor.

With this the following syntax is possible:

```python
t1 = LaguerreGaussianTransverseProfile(20e-6, p=0, m=1)
t2 = LaguerreGaussianTransverseProfile(20e-6, p=1, m=1)
l = GaussianLongitudinalProfile(800e-9, 100e-15, 0)
profile1 = CombinedLongitudinalTransverseProfile(800e-9, (1, 0), 1, l, t1)
profile2 = CombinedLongitudinalTransverseProfile(800e-9, (1, 0), 1, l, t1)

composition = profile1*2+2*profile2
```


RemiLehe:
Thanks for this PR @soerenjalas!
Could you add an automated test, so that this feature is covered by CI?

soerenjalas:
Good point! I added some tests using a mock laser profile that returns constant values. For now I put it in `test_laser_profiles.py`. Maybe it makes sense to put the tests on the base Profile class in a different file to clean things up, though. Let me know what you think.

MaxThevenet:
Thanks for this PR! The GitHub code scanning bot made a few annotations, could you check if they require code changes, or discard them if they aren't relevant?

soerenjalas:
The critical ones about the parent init methods are fixed. The ones with "Statement has no effect" are not relevant in this case  I believe. The function tests whether an error is thrown correctly, therefore by design it shouldn't return.

Is there anything I need to do to acknowledge the CodeQL stuff?

MaxThevenet (approved):
Great, thanks for this PR!

MaxThevenet:
I don't think you need to do anything, there were just a few alerts to dismiss.

PR #111, opened 2023-02-10
Avoid storing `omega` and `field_fft` in the `Laser` class

RemiLehe:
The variables `omega` and `field_fft` are not used anywhere else in the code ; this PR removes them for clean-up.


lasy/laser.py
```diff
@@ -153,17 +150,17 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
                             verbose=False,
                         )
                     )
             # Propagate the spectral image
             for i_m in range(self.box.azimuthal_modes.size):
-                transform_data = np.transpose(self.field.field_fft[i_m]).copy()
+                transform_data = np.transpose(field_fft[i_m]).copy()
                 self.prop[i_m].step(transform_data, distance, overwrite=True)
-                self.field.field_fft[i_m, :, :] = np.transpose(transform_data).copy()
+                field_fft[i_m, :, :] = np.transpose(transform_data).copy()
```
hightower8083, line 163: maybe we could generalize a bit more using  `np.swapaxes(.., time_axis_indx, 0)` instead of `np.transpose`

hightower8083 (approved):
mark a tiny comment above, otherwise looks good to me. by the way, there are still couple of broken things from axes switch, see #113

lasy/laser.py
RemiLehe: Thanks for the suggestion.
Is it okay if I do it in a separate PR, to keep things clean? 
This is because the current PR focuses on removing the attributes `field_fft`,`omega`, etc.

MaxThevenet: I agree, let's keep it separate.

MaxThevenet (approved):
Thanks for this PR!

PR #112, opened 2023-02-10
Add Hermite Gaussian Modal Decomposition Functionality

rob-shalloo:
This PR adds the ability to decompose an Instance of `TransverseProfile` into a series of Hermite Gauss Modes. The projection of the given laser profile onto a user-defined number of Hermite Gauss Modes in performed and in the end a dictionary is provided back to the user. This dictionary contains the mode indices as keys and the weight of that particular mode as the value. During the decomposition, an appropriate spot size must be selected. This is estimated by finding the D4Sigma spot size of the profile and then creating a range of spot sizes close to this value. Ultimately the spot size that is chosen is the spot size from this range which maximises the weight of the first mode.


lasy/utils/exp_data_utils.py
```diff
@@ -18,12 +18,34 @@ def find_center_of_mass(img):
...
     return x0, y0
+
+
+def find_d4sigma(img):
+    """
+    Finds the D4Sigma measurement of the spot size
+    in x and y in pixels.
+
+    https://en.wikipedia.org/wiki/Beam_diameter#D4%CF%83_or_second-moment_width
+
```
MaxThevenet, line 38:
```suggestion

    Parameters
    ----------
    img : A numpy array containing the spatial intensity profile of a laser pulse.

    Returns
    -------
    D4sigX : The D4sigma along the first (x) axis
    D4sigY : The D4sigma along the second (y) axis
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+
+    The function takes either an instance of `TransverseProfile` or an
+    instance of `Laser` (that is, either a transverse profile or the
+    full 3D laser profile defined on a grid). In the case that an
+    instance of `Laser` is passed then the intensity of this profile
+    is projected onto an x-y plane for the decomposition.
+
+    Parameters
+    ----------
+    laserProfile: class instance
```
MaxThevenet: I think this space is there in the numpy docstrings, right?
```suggestion
    laserProfile : class instance
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+        An instance of a class or sub-class of TransverseLaserProfile
+
+    n_x_max, n_y_max: ints
```
MaxThevenet:
```suggestion
    n_x_max, n_y_max : ints
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+        The maximum values of `n_x` and `n_y` out to which the expansion
+        will be performed
+
+    Returns
+    -------
+    weights: dict of floats
+        A dictionary of floats corresponding to the weights of each mode
+        in the decomposition. The keys of the dictionary are tuples
+        corresponding to (`n_x`,`n_y`)
```
MaxThevenet: Is a dictionary sufficient, or should we make it a class with more metadata?

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+    # transverse profile.
+
+    assert isinstance(
+        laserProfile, TransverseProfile
+    ), "laserProfile must be an instance of TransverseProfile"
+
+    # Here we need to define a grid size to use for the calculation
+    # By default we will use N x N grid where N = 500
+    N = 500
+
```
MaxThevenet: Could this be an optional input parameter? Also, this should be mostly used mainly for experimentally measured profiles, right? When that's the case, should we store the original resolution in TransverseProfileFromData, and expose it here, to simplify the use and also avoid additional smoothing?

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+def estimate_best_HG_waist(x, y, field):
+    """
+    Estimate the waist that maximises the weighting of the first mode.
+
+    Calculates a D4Sigma waist as a first estimate and then
+
+
+    Parameters
+    ---------
+    x,y: 1D numpy arrays
```
MaxThevenet:
```suggestion
    x,y : 1D numpy arrays
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+        representing the x and y axes on which the intensity profile is defined.
+
+    field: 2D numpy array
```
MaxThevenet:
```suggestion
    field : 2D numpy array representing the field (not the laser intensity).
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+    coeffTest = np.zeros_like(waistTest)
+
+    for i, wTest in enumerate(waistTest):
+        # create a gaussian
+        HGMode = HermiteGaussianTransverseProfile(wTest, 0, 0)
+        profile = HGMode.evaluate(X, Y)
+        coeffTest[i] = np.sum(profile * field)
+    w0 = waistTest[np.argmax(coeffTest)]
+
+    print(f"Estimated w0 {w0Est*1e6}")
```
MaxThevenet: Or did I misunderstand?
```suggestion
    print(f"Estimated w0 {w0Est*1e6} microns")
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return weights
+
+
+def estimate_best_HG_waist(x, y, field):
+    """
+    Estimate the waist that maximises the weighting of the first mode.
+
+    Calculates a D4Sigma waist as a first estimate and then
+
```
MaxThevenet:
```suggestion
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+            weights[(i, j)] = coef
+
+    return weights
+
+
+def estimate_best_HG_waist(x, y, field):
+    """
+    Estimate the waist that maximises the weighting of the first mode.
+
+    Calculates a D4Sigma waist as a first estimate and then
```
MaxThevenet: Isn't the end of this sentence missing?

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+    print(D4SigX)
+    print(D4SigY)
+    w0Est = np.mean((D4SigX, D4SigY)) / 2 * dx  # convert this to a 1/e^2 width
+
+    waistTest = np.linspace(w0Est / 2, w0Est * 1.5, 15)
+    coeffTest = np.zeros_like(waistTest)
+
+    for i, wTest in enumerate(waistTest):
+        # create a gaussian
+        HGMode = HermiteGaussianTransverseProfile(wTest, 0, 0)
```
MaxThevenet, line 140: Having a profile normalised to the energy is required here, and this is why you changed the`evaluate` function of `HermiteGaussianTransverseProfile`, right?

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,133 @@
...
+    assert dx == dy
+
+    X, Y = np.meshgrid(x, y)
+
+    D4SigX, D4SigY = find_d4sigma(np.abs(field) ** 2)
+    print(D4SigX)
+    print(D4SigY)
+    w0Est = np.mean((D4SigX, D4SigY)) / 2 * dx  # convert this to a 1/e^2 width
+
+    waistTest = np.linspace(w0Est / 2, w0Est * 1.5, 15)
```
MaxThevenet:
```suggestion
    # Scan around the waist obtained from the D4sigma calculation,
    # and keep the waist for which this HG mode has the highest scalar
    # product with the input profile.
    waistTest = np.linspace(w0Est / 2, w0Est * 1.5, 15)
```

MaxThevenet (commented):
Great, thanks for this PR! Could you add a CI test? Also (and this may have been discussed), it would be nice to have a function to generate the sum of HG profiles from the dictionaries. The recent PR by @soerenjalas should make this reasonably easy. The CI test could then be to decompose an arbitrary profile, get the HG coefficients, generate the profile consisting in the sum of HG modes, and make sure that the difference between original profile and the sum of HG modes is a small fraction of the original profile. This may be a bit overkill for a test, but is probably something that users would naturally turn to, what do you think?

rob-shalloo:
sampleSpot

rob-shalloo:
sampleSpot_bgSub
Image to be used in the example of the decomposition/reconstruction of experimental data

lasy/utils/mode_decomposition.py
rob-shalloo: I think the point on oversmoothing is a good one, although I think we should leave it to the user to decide the value of N. I will expose as an input parameter with a default of 500 and leave it to the user

lasy/utils/mode_decomposition.py
rob-shalloo: yeah, good catch thanks. Fixed

lasy/utils/mode_decomposition.py
rob-shalloo: a class may be a better solution in the long run, but I would suggest that for the moment we proceed like this and then make a change once we have a better idea of how we're using it

lasy/utils/mode_decomposition.py
rob-shalloo: a normalised profile is not technically required here. As long as the calling of `HermiteGaussianTransverseProfile` is consistent, then the numbers will work out to give the right intensity. I changed the evaluate part of `HermiteGaussianTransverseProfile` so that all modes have the same 2D integral because I think it's tidier and I think cleaner/easier to understand.

rob-shalloo:
Example of reconstruction of supergaussian
supergauss_reconstruction

rob-shalloo:
Example of reconstruction of a spot which is imported from data
data_reconstruction

lasy/utils/mode_decomposition.py
MaxThevenet: Fine with me

MaxThevenet:
Works like a charm! A few points that we should discuss offline:
- Let's see if we can simplify CI, I can help.
- Would be nice to merge the two examples, if they have a lot of duplication.
- Some changes appear that should probably not be there. Could you merge dev into your branch?
- Could you use COM capability implemented in lasy/utils/exp_data_utils.py, and/or improve it?

rob-shalloo:
Have updated code to remove duplicate example and to move centering functions into the`TransverseProfileFromData` class and implemented it using the existing `find_center_of_mass` function

MaxThevenet:
Thanks! There are a few comments from CodeQL, could you check if they are valid and, if so, fix the code? Then you can re-request a review from me whenever the PR is ready.

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+
+import matplotlib.pyplot as plt
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+import skimage
+
+
+# Define the transverse profile of the laser pulse
+img_url = "https://user-images.githubusercontent.com/27694869/228038930-d6ab03b1-a726-4b41-a378-5f4a83dc3778.png"
+intensityData = skimage.io.imread(img_url)
+rows, cols = intensityData.shape
```
MaxThevenet:
```suggestion
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+intensityData[intensityData < 2.1] = 0
+pixel_calib = 0.186e-6
+lo = (
+    -intensityData.shape[0] / 2 * pixel_calib,
+    -intensityData.shape[0] / 2 * pixel_calib,
+)
+hi = (
+    intensityData.shape[0] / 2 * pixel_calib,
+    intensityData.shape[0] / 2 * pixel_calib,
+)
```
MaxThevenet: ? Are we just fortunate that it's a square image?
```suggestion
lo = (
    -intensityData.shape[0] / 2 * pixel_calib,
    -intensityData.shape[1] / 2 * pixel_calib,
)
hi = (
    intensityData.shape[0] / 2 * pixel_calib,
    intensityData.shape[1] / 2 * pixel_calib,
)
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Combine into full laser profile
+profile = CombinedLongitudinalTransverseProfile(
+    wavelength, pol, energy, longProfile, transProf
+)
+
+# Calculate the decomposition into hermite-gauss modes
+n_x_max = 20
+n_y_max = 20
+modeCoeffs, wasit = hermite_gauss_decomposition(
```
MaxThevenet:
```suggestion
modeCoeffs, waist = hermite_gauss_decomposition(
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+    -intensityData.shape[0] / 2 * pixel_calib,
+    -intensityData.shape[0] / 2 * pixel_calib,
+)
+hi = (
+    intensityData.shape[0] / 2 * pixel_calib,
+    intensityData.shape[0] / 2 * pixel_calib,
+)
+energy = 0.5
+pol = (1, 0)
+transProf = TransverseProfileFromData(intensityData, lo, hi)
```
MaxThevenet: Name inconsistent with `longProfile`. Also, we probably should enforce a convention with snake_case for objects and CamelCase for classes.
```suggestion
transverse_profile = TransverseProfileFromData(intensityData, lo, hi)
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Define longtiudinal profile of the laser pulse
+wavelength = 800e-9
+tau = 30e-15
+t_peak = 0.0
+longProfile = GaussianLongitudinalProfile(wavelength, tau, t_peak)
```
MaxThevenet:
```suggestion
longitudinal_profile = GaussianLongitudinalProfile(wavelength, tau, t_peak)
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Combine into full laser profile
+profile = CombinedLongitudinalTransverseProfile(
+    wavelength, pol, energy, longProfile, transProf
```
MaxThevenet:
```suggestion
    wavelength, pol, energy, longitudinal_profile, transverse_profile
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+)
+
+# Calculate the decomposition into hermite-gauss modes
+n_x_max = 20
+n_y_max = 20
+modeCoeffs, wasit = hermite_gauss_decomposition(
+    transProf, n_x_max=n_x_max, n_y_max=n_y_max, N_pts=500
```
MaxThevenet:
```suggestion
    transverse_profile, n_x_max=n_x_max, n_y_max=n_y_max, N_pts=500
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+)
+
+# Reconstruct the pulse using a series of hermite-gauss modes
+for i, mode_key in enumerate(list(modeCoeffs)):
+    tmpTranProf = HermiteGaussianTransverseProfile(wasit, mode_key[0], mode_key[1])
```
MaxThevenet:
```suggestion
    tmp_transverse_profile = HermiteGaussianTransverseProfile(waist, mode_key[0], mode_key[1])
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+    if i == 0:
+        reconstructedProfile = modeCoeffs[
+            mode_key
+        ] * CombinedLongitudinalTransverseProfile(
+            wavelength, pol, energy, longProfile, tmpTranProf
```
MaxThevenet:
```suggestion
            wavelength, pol, energy, longitudinal_profile, tmp_transverse_profile
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+    else:
+        reconstructedProfile += modeCoeffs[
+            mode_key
+        ] * CombinedLongitudinalTransverseProfile(
+            wavelength, pol, energy, longProfile, tmpTranProf
```
MaxThevenet:
```suggestion
            wavelength, pol, energy, longitudinal_profile, tmp_transverse_profile
```

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+# Plotting the results
+x = np.linspace(-5 * wasit, 5 * wasit, 500)
```
MaxThevenet:
```suggestion
x = np.linspace(-5 * waist, 5 * waist, 500)
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,143 @@
...
+        The number of grid points in x and y that will be used during the
+        decomposition calculation
+
+    Returns
+    -------
+    weights: dict of floats
+        A dictionary of floats corresponding to the weights of each mode
+        in the decomposition. The keys of the dictionary are tuples
+        corresponding to (`n_x`,`n_y`)
+
```
MaxThevenet, line 43:
```suggestion
        corresponding to (`n_x`,`n_y`)

        waist : Beam waist for which the decomposition is calculated.
            It is computed as the waist for which the weight of order 0 is maximum. 
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,143 @@
...
+        The maximum values of `n_x` and `n_y` out to which the expansion
+        will be performed
+
+    N_pts : int
+        The number of grid points in x and y that will be used during the
+        decomposition calculation
+
+    Returns
+    -------
+    weights: dict of floats
```
MaxThevenet:
```suggestion
    weights : dict of floats
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,143 @@
...
+from lasy.utils.exp_data_utils import find_d4sigma
+
+import numpy as np
+import math
+
+
+def hermite_gauss_decomposition(laserProfile, n_x_max=12, n_y_max=12, N_pts=500):
+    """
+    Decomposes a `lasy` laser profile into a set of hermite-gaussian
+    modes.
```
MaxThevenet:
```suggestion
    Decomposes a laser profile into a set of hermite-gaussian modes.
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,143 @@
...
+    field : 2D numpy array representing the field (not the laser intensity).
+        the laser field profile in a 2D slice.
+
+    """
+
+    dx = x[2] - x[1]
+    dy = y[2] - y[1]
+
+    assert dx == dy
+
```
MaxThevenet, line 126: Maybe this is required? I didn't test this assert
```suggestion
    assert field.shape == (x.size, y.size)
```

MaxThevenet (commented):

lasy/utils/mode_decomposition.py
rob-shalloo: I don't think it's necessary for x and y to be the same length, in fact for most experimental data this won't be the case, cameras tend to have rectangular chips. But I do think we start to have problems if dx and dy are not the same

MaxThevenet: That's not testing `x.size == y.size`.

examples/example_modal_decomposition_data.py
rob-shalloo: nope, that was a mistake

lasy/utils/mode_decomposition.py
rob-shalloo: yes, i misunderstood the original point. I think this makes sense

rob-shalloo:
so looking at it again, I think another asset is needed or possibly a slight change to the start of `hermite_gauss_decomposition`. With the catching of the error above for 
```
lo = (
    -intensityData.shape[0] / 2 * pixel_calib,
    -intensityData.shape[1] / 2 * pixel_calib,
)
hi = (
    intensityData.shape[0] / 2 * pixel_calib,
    intensityData.shape[1] / 2 * pixel_calib,
)
```
we now have an issue when we define the field as an NxN grid that 
`assert dx == dy`
fails. This makes sense. I will address soon.

rob-shalloo:
fixed

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,112 @@
...
+from lasy.profiles.transverse.hermite_gaussian_profile import (
+    HermiteGaussianTransverseProfile,
+)
+from lasy.profiles.longitudinal.gaussian_profile import GaussianLongitudinalProfile
+from lasy.profiles.combined_profile import CombinedLongitudinalTransverseProfile
+from lasy.utils.mode_decomposition import hermite_gauss_decomposition
+
+import matplotlib.pyplot as plt
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+import skimage
```
MaxThevenet, line 15: Just a note: if we test this script in CI, we'll need to install a few more libraries. Maybe we should have an `examples/requirements.txt`? This is, however, not required for this PR.

examples/example_modal_decomposition_data.py
```diff
@@ -0,0 +1,112 @@
...
+)
+hi = (
+    intensityData.shape[0] / 2 * pixel_calib,
+    intensityData.shape[1] / 2 * pixel_calib,
+)
+energy = 0.5
+pol = (1, 0)
+transverse_profile = TransverseProfileFromData(intensityData, lo, hi)
+
+# Define longtiudinal profile of the laser pulse
```
MaxThevenet:
```suggestion
# Define longitudinal profile of the laser pulse
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns
+    -------
+    weights : dict of floats
+        A dictionary of floats corresponding to the weights of each mode
+        in the decomposition. The keys of the dictionary are tuples
+        corresponding to (`n_x`,`n_y`)
+
+        waist : Beam waist for which the decomposition is calculated.
+            It is computed as the waist for which the weight of order 0 is maximum.
```
MaxThevenet: That's probably my bad, I may have messed up with the indentation in my comment.
```suggestion
    waist : Beam waist for which the decomposition is calculated.
        It is computed as the waist for which the weight of order 0 is maximum.
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Parameters
+    ---------
+    x,y : 1D numpy arrays
+        representing the x and y axes on which the intensity profile is defined.
+
+    field : 2D numpy array representing the field (not the laser intensity).
+        the laser field profile in a 2D slice.
+
+    """
```
MaxThevenet, line 121:
```suggestion
    Returns
    -------
    w0 : scalar
        The calculated waist.
    """
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,144 @@
...
+
+    dx = x[2] - x[1]
+    dy = y[2] - y[1]
+    print(dx)
+    print(dy)
```
MaxThevenet:
```suggestion
```

lasy/utils/mode_decomposition.py
```diff
@@ -0,0 +1,144 @@
...
+    x,y : 1D numpy arrays
+        representing the x and y axes on which the intensity profile is defined.
+
+    field : 2D numpy array representing the field (not the laser intensity).
+        the laser field profile in a 2D slice.
+
+    """
+
+    dx = x[2] - x[1]
+    dy = y[2] - y[1]
```
MaxThevenet: Is this equivalent? Would help someone working with very small arrays...
```suggestion
    dx = x[1] - x[0]
    dy = y[1] - y[0]
```

MaxThevenet (commented):
Great, thanks! I believe the remaining comments are pretty small, see below.

MaxThevenet (approved):
Thanks for this PR!

PR #113, opened 2023-02-11
fix propagator absorption get_full_field with respect to new time axis

hightower8083:
fixed remaining effects of t-axis swap in the propagator absorbtion and get_full_field


MaxThevenet (approved):
Thanks for the fixes!

PR #114, opened 2023-02-11
Add time axis refinements to get_full_field method

hightower8083:
In case if we wish to keep get_full_field method for some more time, here's a version with time-axis refinement, so no need to use high resolution in time for making plots
Should probably be merged after #113

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -1,5 +1,6 @@
 import numpy as np
 import scipy.constants as scc
+from scipy.interpolate import interp1d
 from axiprop.lib import PropagatorFFT2, PropagatorResampling
 
 from lasy.utils.box import Box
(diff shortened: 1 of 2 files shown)


MaxThevenet:
As discussed in the last meeting, IMO this feature probably doesn't need to be in the core of LASY, which deals with the envelope:
- The slicing should rather be a feature of openPMD-viewer (it already is, I have not checked but it should work with lasy data)
- The conversion from envelope to field is given by `field = Re(envelope * exp(-1j*omega0*t))`
- As long as complex interpolation gives nice results (and we therefore don't need more advanced FFT-based methods), increasing the resolution is a relatively simple operation.

Therefore, I would remove this feature from the `Laser` class. We can keep it in `utils` in a separate function.

Also, I think all relevant PRs are merged, you can probably merge `development` into your branch, which should remove some previous changes.

hightower8083:
moved this function to `utils.laser_utils` and merged the development. 
should be fine to merge this PR

MaxThevenet (approved):
Thanks for this PR!

PR #118, opened 2023-03-13
Update openPMD metadata

RemiLehe:
As discussed in #22 the box cell size is actually not `(hi-lo)/npoints` but `(hi-lo)/(npoints-1)`.


MaxThevenet (approved):
Thanks for this PR. I agree, let's make it consistent now, and then we can change it if it seems more appropriate.

PR #121, opened 2023-04-03
Add option to initialize Gaussian pulse out of focus

RemiLehe:
Add the option to initialize a Gaussian pulse out of focus, using analytical formulas from the paraxial approximation.

If we agree on the name of the argument and the corresponding documentation, I'll add this for other pulses for which an analytical formula is available.

### DIFF
--- a/lasy/profiles/transverse/gaussian_profile.py
+++ b/lasy/profiles/transverse/gaussian_profile.py
@@ -7,7 +7,7 @@ class GaussianTransverseProfile(TransverseProfile):
     """
     Derived class for the analytic profile of a Gaussian laser pulse.
 
-    More precisely, the transverse envelope
+    More precisely, at focus (`z_init=0`), the transverse envelope
     (to be used in the :class:CombinedLongitudinalTransverseLaser class)
     corresponds to:
 
@@ -19,11 +19,20 @@ class GaussianTransverseProfile(TransverseProfile):
     ----------
     w0 : float (in meter)
         The waist of the laser pulse, i.e. :math:`w_0` in the above formula.
+
+    wavelength : float (in meter)
+        The main laser wavelength :math:`\\lambda_0` of the laser.
+
+    z_init : float (in meter), optional
+        Position at which the pulse is initialized, relative to the position
+        of the focal plane. (The focal plane is at `z=0`.)
     """
 
-    def __init__(self, w0):
+    def __init__(self, w0, wavelength, z_init=0):
         super().__init__()
         self.w0 = w0
+        self.inv_zr = wavelength/(np.pi*w0**2)
+        self.z_init = z_init
 
     def _evaluate(self, x, y):
         """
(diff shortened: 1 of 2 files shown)


lasy/profiles/transverse/gaussian_profile.py
```diff
@@ -19,11 +19,28 @@ class GaussianTransverseProfile(TransverseProfile):
...
+        In order to initialize the pulse out of focus, you can either:
+            - Use a non-zero `z_init`
+            - Use `z_init=0` and then call `laser.propagate(z_init)`
+        Both methods are in principle equivalent, but note that the first
+        method uses the paraxial approximation, while the second method does
+        not make this approximation.
     """
 
-    def __init__(self, w0):
+    def __init__(self, w0, wavelength, z_init=0):
```
MaxThevenet: Could we make `wavelength` an optional parameter (mandatory if z_init is not 0`)? None of the other transverse profile require the wavelength.

lasy/profiles/transverse/gaussian_profile.py
```diff
@@ -19,11 +19,28 @@ class GaussianTransverseProfile(TransverseProfile):
     ----------
     w0 : float (in meter)
         The waist of the laser pulse, i.e. :math:`w_0` in the above formula.
+
+    wavelength : float (in meter)
+        The main laser wavelength :math:`\\lambda_0` of the laser.
+
+    z_init : float (in meter), optional
```
MaxThevenet: This name sounds good to me, as would the now traditional `z_foc`. Any other opinion?

MaxThevenet (commented):
Looks good to me, thanks! I don't have a strong opinion on the parameter name, see a few comments below.

lasy/profiles/transverse/gaussian_profile.py
RemiLehe: Discussed offline: yes, we will change this to `z_foc`, and mention that the equivalent `propagate` call would be `.propagate(-z_foc)`

lasy/profiles/transverse/gaussian_profile.py
```diff
@@ -19,11 +19,28 @@ class GaussianTransverseProfile(TransverseProfile):
...
+        The main laser wavelength :math:`\\lambda_0` of the laser.
+
+    z_foc : float (in meter), optional
+        Position of the focal plane. (The laser pulse is initialized at `z=0`.)
+
+    .. warning::
+
+        In order to initialize the pulse out of focus, you can either:
+            - Use a non-zero `z_foc`
+            - Use `z_foc=0` (i.e. initialize the pulse at focuse) and then
```
MaxThevenet:
```suggestion
            - Use `z_foc=0` (i.e. initialize the pulse at focus) and then
```

lasy/profiles/transverse/gaussian_profile.py
```diff
@@ -19,11 +19,28 @@ class GaussianTransverseProfile(TransverseProfile):
...
+              call `laser.propagate(-z_foc)`
+        Both methods are in principle equivalent, but note that the first
+        method uses the paraxial approximation, while the second method does
+        not make this approximation.
     """
 
-    def __init__(self, w0):
+    def __init__(self, w0, wavelength, z_foc=0):
```
MaxThevenet, line 40: Should we make `wavelength` optional and check that it is mentioned only if `z_foc` is specified (or non-zero)? I believe this could help consistency with other profiles. No strong opinion.

MaxThevenet (commented):
Thanks for updating this PR! See a few points below.

lasy/profiles/transverse/gaussian_profile.py
MaxThevenet: As discussed offline, let's merge as is and fix this later.

MaxThevenet (approved):

PR #123, opened 2023-04-03
Fix ordering of `globalOffset` in openPMD metadata

RemiLehe:
I forgot to switch the ordering of `globalOffset` in this PR: https://github.com/LASY-org/lasy/pull/105


MaxThevenet:
This fixes the issue discussed offline indeed, thanks!

MaxThevenet (approved):

PR #126, opened 2023-04-21
Implement methods for conversion between temporal and spatial field presentations

hightower8083:
This PR adds to main `Laser` class the  methods `export_to_z` and `import_from_z` for conversion between temporal and spatial field presentations, and the test script for these methods. 

Additionally this PR includes:
- fix of the use of `fft` method in  propagator to comply with the used envelope definition: `ifft` with `"backward"` normalization should be used to convert the field to the frequency domain
- direct import of `c` from `scipy.constants` is using instead of `scc.c`
- removed the requirement for the specific version of `axiprop`


lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+        """
+        Convert laser pulse defined in the temporal domain to the spatial domain
+
+        Parameters
+        ----------
+        z_axis : 1D ndarray of doubles (optional)
+            Spatial `z` axis along which the field should be reconstructed.
+            If not provided, `z_axis = c * t_axis` is considered.
+
+        z0: scalar (optional)
```
MaxThevenet:
```suggestion
        z0 : scalar (optional)
```

lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            Position from which the field is produced (emitted)
+
+        t0: scalar (optional)
```
MaxThevenet:
```suggestion
        t0 : scalar (optional)
```

lasy/laser.py
```diff
@@ -178,17 +180,17 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
                 ]
 
         # Transform the field from temporal to frequency domain
-        field_fft = np.fft.fft(self.field.field, axis=time_axis_indx, norm="forward")
+        field_fft = np.fft.ifft(self.field.field, axis=time_axis_indx, norm="backward")
```
MaxThevenet, line 189: Was this a bug in the previous version? Could you comment on it?

lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+        Convert laser pulse defined in the temporal domain to the spatial domain
+
+        Parameters
+        ----------
+        z_axis : 1D ndarray of doubles (optional)
+            Spatial `z` axis along which the field should be reconstructed.
+            If not provided, `z_axis = c * t_axis` is considered.
+
+        z0: scalar (optional)
+            Position from which the field is produced (emitted)
```
MaxThevenet, line 266: Is this the z0 position in `E(x,y,z0,t)` for the current representation of the field?

lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+
+        t0: scalar (optional)
+            Moment of time at which the field is produced
```
MaxThevenet, line 269: Is this the t0 position in E(x,y,z,t0) for the returned representation of the field?

lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+    def import_from_z(self, field_z, z_axis, z0=0.0, t0=0.0, backend="NP"):
+        """
+        Export laser pulse from the field map in the spatial domain
+
+        Parameters
+        ----------
+        z_axis : 1D ndarray of doubles
+            Spatial `z` axis along which the field should be reconstructed.
+
+        z0: scalar (optional)
```
MaxThevenet:
```suggestion
        z0 : scalar (optional)
```

lasy/laser.py
```diff
@@ -239,12 +239,162 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            Position at which the field should be recorded
+
+        t0: scalar (optional)
```
MaxThevenet:
```suggestion
        t0 : scalar (optional)
```

MaxThevenet (commented):
Thanks for this PR!! See comments below. Furthermore, could you add a CI test? Suggestions:
- in 3D, for the Gaussian example that you shared in your figures, generate a beam with lasy (so, time representation), call `export_to_z`, and compare with theory (the tolerance will ofc depend on the resolution, but it should be tight enough for the t representation to largely fail the test). Then you could call `import_from_z`, and compare with the original profile.
- Same test in RZ, as these use quite different functions.

lasy/laser.py
hightower8083: actually i'm unsure -- this was clearly an inconsistency in the definition, but it appeared compensated for the envelope propagator, i'm not sure if it was actually correct, but it looked fine in the tests.. I've discovered this when did conversions so made them all consistent.

lasy/laser.py
hightower8083: basically we calculate on `z_axis` the field that is emitted by an antenna placed at `z0` for the time axis as defined by `laser` object, and `t0` allows to additionally adjust time moment corresponding to the reconstruction. can be needed if you have measured the field at focal plane but wanna input it out of focus

MaxThevenet:
Could you have a look at the pre-commit.ci issue?

MaxThevenet:
Thanks! as described in the message above, could you add a CI test for this PR? This should be consistent with what was discussed also with @RemiLehe at a LASY meeting a few weeks back.

hightower8083:
actually i could not yet figure whats wrong with pre-commit. the error is 
```
lasy/laser.py:249 in public method `export_to_z`:
        D400: First line should end with a period (not 'n')
lasy/laser.py:327 in public method `import_from_z`:
        D400: First line should end with a period (not 'n')
```
so it indeed concerns only doc formatting of new functions  but it makes no sense to me. internet says it may relate to the absence of a `.`  at the end of the first line, but its same in `propagate` and it doesn't raise any error..

hightower8083:
@MaxThevenet, @RemiLehe  

I've added a simple test where we take field in `t` domain, export it to `z` domain and compare (with looser  tolerance) with the original field considering approximation`t=-z/c` (for this I consider a large beam), then transform it back to `t` domain, and compare it to the original field (with higher accuracy).

I've also used for convenience a new `axiprop` container class which handles spectral-temporal transforms, axes and boundaries. This makes code clearer, and i'm thinking of passing to this approach in the rest of `axiprop`-related methods. Please take a look at the code in `export_to_z` and tell me what you think.

I've also removed forced version from `axiprop` in **pip** setup

tests/test_t2z2t.py
```diff
@@ -0,0 +1,68 @@
...
+
+    ind0 = laser_t_in.field.field.shape[0] // 2 - 1
+
+    Field_orig = np.real(laser_t_in.field.field[ind0] * np.exp(1j * env_phase_t))
+    Field_approx = np.real(laser_z[ind0] * np.exp(1j * env_phase_z))[:, ::-1]
+
+    assert np.allclose(
+        laser_t_in.field.field[ind0], laser_t_out.field.field[ind0], atol=1e-8
+    )
+    assert np.allclose(Field_approx, Field_orig, atol=2e-3)
```
MaxThevenet: Are `Field_orig` and `Field_approx` supposed to be close to each other in general? Otherwise, could we compare instead something like (pseudo-code) `laser_t_in` and `laser_z*(1+z**2/zR**2)**.5` for a beam with a length = zR centered around t=0 and z=0? These should be close.

MaxThevenet (commented):
Thanks for adding the test! See comment below.

hightower8083:
Hey @MaxThevenet , I've revised the test
Now operates with a case with strong difference with `t` and `z` representations. It  verify the _reversibility_ of `t-z-t` transform and also compares the obtained `z` representation with theory. 
On my machine tests runs fine, but for some reason fails in CI and I fail to understand the reason. Could you check how it runs on your side and let me know if you have any idea how to debug the unittest ?

lasy/laser.py
```diff
@@ -243,12 +245,160 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            field_fft, axis=time_axis_indx, norm="backward"
         )
 
         # Translate phase of the retrieved envelope by the distance
-        self.field.field *= np.exp(1j * self.profile.omega0 * distance / scc.c)
+        self.field.field *= np.exp(1j * omega0 * distance / c)
+
+    def export_to_z(self, z_axis=None, z0=0.0, t0=0.0, backend="NP"):
+        """
+        Convert laser pulse defined in the temporal domain to the spatial domain.
```
MaxThevenet:
```suggestion
        Export laser pulse defined in the temporal domain (internal LASY representation) to the spatial domain.
```

lasy/laser.py
```diff
@@ -243,12 +245,160 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            # Convert the spectral image to the spatial field representation
+            FieldAxprp.import_field(np.transpose(self.field.field).copy())
+            field_z = prop.t2z(FieldAxprp.Field_ft, z_axis, z0=z0, t0=t0).T
+            field_z *= np.exp(-1j * (z_axis / c + t0) * omega0)
+
+        return field_z
+
+    def import_from_z(self, field_z, z_axis, z0=0.0, t0=0.0, backend="NP"):
+        """
+        Export laser pulse from the field map in the spatial domain.
```
MaxThevenet:
```suggestion
        Import laser pulse defined in the spatial domain to the temporal domain (internal LASY representation).
```

lasy/laser.py
```diff
@@ -243,12 +245,160 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            # Convert the spectral image to the spatial field representation
+            FieldAxprp.import_field(np.transpose(self.field.field).copy())
+            field_z = prop.t2z(FieldAxprp.Field_ft, z_axis, z0=z0, t0=t0).T
+            field_z *= np.exp(-1j * (z_axis / c + t0) * omega0)
+
+        return field_z
+
+    def import_from_z(self, field_z, z_axis, z0=0.0, t0=0.0, backend="NP"):
+        """
+        Import laser pulse defined in the spatial domain to the temporal domain (internal LASY representation).
```
MaxThevenet:
```suggestion
        Import laser pulse defined in the spatial domain to the
        temporal domain (internal LASY representation).
```

lasy/laser.py
```diff
@@ -243,12 +245,160 @@ def propagate(self, distance, nr_boundary=None, backend="NP"):
...
+            field_fft, axis=time_axis_indx, norm="backward"
         )
 
         # Translate phase of the retrieved envelope by the distance
-        self.field.field *= np.exp(1j * self.profile.omega0 * distance / scc.c)
+        self.field.field *= np.exp(1j * omega0 * distance / c)
+
+    def export_to_z(self, z_axis=None, z0=0.0, t0=0.0, backend="NP"):
+        """
+        Export laser pulse defined in the temporal domain (internal LASY representation) to the spatial domain.
```
MaxThevenet:
```suggestion
        Export laser pulse defined in the temporal domain
        (internal LASY representation) to the spatial domain.
```

tests/test_t2z2t.py
```diff
@@ -0,0 +1,107 @@
...
+    omega0 = profile.long_profile.omega0
+    k0 = omega0 / c
+    lambda0 = 2 * np.pi / k0
+
+    L_Ray = np.pi * w0**2 / lambda0
+    z_axis_2d = z_axis[None, :]
+    r_axis_2d = r_axis[:, None]
+    w0_z = w0 * np.sqrt(1 + (z_axis_2d / L_Ray) ** 2)
+    R_z_inv = z_axis_2d / (z_axis_2d**2 + L_Ray**2)
+    phi_goy = np.arctan2(z_axis_2d, L_Ray)
```
MaxThevenet:
```suggestion
    phi_gouy = np.arctan2(z_axis_2d, L_Ray)
```

tests/test_t2z2t.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Field = (
+        w0
+        / w0_z
+        * np.exp(-(r_axis_2d**2) / w0_z**2)
+        * np.exp(-(z_axis_2d**2) / (c * tau) ** 2)
+        * np.exp(1j * (k0 * r_axis_2d**2 * R_z_inv / 2 - phi_goy))
```
MaxThevenet:
```suggestion
        * np.exp(1j * (k0 * r_axis_2d**2 * R_z_inv / 2 - phi_gouy))
```

MaxThevenet (commented):
Looks great, thanks for this PR! Below are minor typos fixes.

MaxThevenet (approved):

PR #133, opened 2023-04-27
Set Version in Documentation

RemiLehe:


docs/source/conf.py
```diff
@@ -24,7 +24,7 @@
 author = "LASY-org"
 
 # The full version, including alpha/beta/rc tags
-release = "0.0"
+release = "0.1.0"
```
AngelFP: I would suggest something like
```python
from lasy import __version__

release = __version__
```
so that we have a single source for the version number (this is what we do in optimas, for example).

docs/source/conf.py
```diff
@@ -13,6 +13,7 @@
 import os
 import sys
 from datetime import date
+from lasy import __version__
```
ax3l, line 16: Extra space to trigger CI
```suggestion
from lasy import __version__

```

ax3l:
For this to build, we need to properly install lazy probably for RTD, instead of side-channeling the include path.

We do something like this here for the `Python/` directory:
https://github.com/ECP-WarpX/WarpX/blob/63f8f713e2ea37c63b5c5ca82beb41683dd0b7ce/Docs/requirements.txt#L8

docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
--r ../requirements.txt
+-e ..
```
ax3l: Actually, it builds from the source root
```suggestion
# install this from the source root (!) directory like this:
#   python -m pip install --upgrade -r docs/requirements.txt
-e .
```

ax3l: Actually, it builds from the source root
```suggestion
# install this from the source root (!) directory like this:
#   python -m pip install --upgrade -r docs/requirements.txt
-e .
```

ax3l (approved):

Issue #132, opened 2023-04-27
Depend on a tagged/release axiprop


ax3l:
We cannot release lasy to PyPI until axiprop is also uploaded:
https://github.com/hightower8083/axiprop/issues/8

current zip:
```
Uploading lasy-0.1.0-py3-none-any.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 40.7/40.7 kB • 00:00 • 48.9 MB/s
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Invalid value for requires_dist. Error: Can't have direct dependency:  
         'axiprop @                                                             
         https://github.com/hightower8083/axiprop/archive/refs/heads/main.zip'  
```

same via git
```
Uploading lasy-0.1.0-py3-none-any.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 40.7/40.7 kB • 00:00 • 45.7 MB/s
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Invalid value for requires_dist. Error: Can't have direct dependency:  
         'axiprop @                                                             
         git+https://github.com/hightower8083/axiprop.git@9bd516898c1cc4e1bcbfb9
         adac26f8909ec087a0'                                                    
```

ax3l:
Fixed via #135

PR #135, opened 2023-05-09
set axiprop to be installed from pip

hightower8083:
that should address #132


ax3l (approved):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-axiprop @ https://github.com/hightower8083/axiprop/archive/refs/heads/main.zip
+axiprop==0.5
```
ax3l, line 1: If you like, we can also keep the version open (at latest) and fix breaking changes as the occur :)

hightower8083: in principle sure, but let me handle this when breaking change occurs -- for now let's keep it fixed against any stupid actions from me =) this version it already supports #126, so next time i'll update if we decide to enable direct propagation of z-defined field

PR #138, opened 2023-05-17
Fix user guide

RemiLehe:
The code in the user guide does not run as is, due to missing import statements. This PR fixes it.
Eventually, it would be good to run the documentation examples as part of our CI.


rob-shalloo (approved):
Hey Remi, thanks for fixing this!

PR #140, opened 2023-05-18
Store envelope openPMD metadata at the right level

RemiLehe:
According to the [openPMD standard](https://github.com/MaxThevenet/openPMD-standard/blob/0d8fb0d8722c8a0b8f80967cd83dd022eaa2a30f/EXT_LaserEnvelope.md), the laser information (e.g. polarization, frequency) should be stored at the mesh level.


MaxThevenet (approved):
Thanks for fixing this! We could briefly chat what makes more sense w.r.t. common usage open openPMD-standard, but for now let's just make sure this is consistent.

PR #141, opened 2023-05-20
New profile: from numpy array

MaxThevenet:
This PR proposes a new Profile, created from a numpy array. With the following script, it makes it possible to read the output of an FBPIC simulation, and use it as an input in a HiPACE++ simulation, as can be seen in the figure below.
fig

It uses `scipy.interpolate` to enable a different resolution in the FBPIC output and in the LASY representation.

The final implementation could have a new profile `FromSimulationProfile` deriving from `FromArrayProfile` (so the option to initialize a profile from a numpy array is easily exposed). With this approach, the role of each class would be:
- `FromSimulationProfile`: Read simulation output, convert to envelope representation, handle polarization etc. Could use openpmd-viewer.
- `FromArrayProfile`: convert between geometries if needed (this could even be in the class above, not sure yet). Would not use openpmd-viewer.

```py
import numpy as np
from scipy.constants import c
from openpmd_viewer import OpenPMDTimeSeries
from scipy.signal import hilbert
from lasy.laser import Laser
from lasy.profiles.from_array_profile import FromArrayProfile

wavelength = 0.8e-6

# Read output of a FBPIC simulation, extract to 3D for simplicity
ts = OpenPMDTimeSeries('lab_diags/hdf5/')
F, m = ts.get_field(iteration=10, field='E', coord='y', theta=None)

# Get envelope from full profile
k0 = 2*np.pi/wavelength
env_F = hilbert(F)*np.exp(-1j*k0*m.z)

# Create LASY profile
pol = (1,0)
dim = 'xyt'
array = np.flip(env_F, axis=2) # Need to flip z vs. t
axes = {'x':m.x, 'y':m.y, 't':(m.z-m.z[0])/scc.c}
profile = FromArrayProfile(wavelength=wavelength, pol=pol, array=array, axes=axes, dim=dim)

# Create LASY laser object and save to file
lo = (-120e-6, -120e-6, -10e-15)
hi = (+120e-6, +120e-6, +60e-15)
npoints=(32,64,128)
laser = Laser(dim, lo, hi, npoints, profile)
laser.write_to_file('./laserfbpic/laser3d')
```

### DIFF
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -34,7 +34,7 @@ class FromArrayProfile(Profile):
         ['x', 'y', 'z'] and ['z', 'y', 'x'].
     """
 
-    def __init__(self, wavelength, pol, array, axes, dim, axes_order=['x', 'y', 'z']):
+    def __init__(self, wavelength, pol, array, axes, dim, axes_order=["x", "y", "z"]):
         super().__init__(wavelength, pol)
 
         assert dim == "xyt" or dim == "rt", "dim must be 'xyt' or 'rt'"
@@ -44,10 +44,9 @@ class FromArrayProfile(Profile):
         self.dim = dim
         self.axes = axes
         if dim == "xyt":
+            assert axes_order in [["x", "y", "z"], ["z", "y", "x"]]
 
-            assert axes_order in [['x', 'y', 'z'], ['z', 'y', 'x']]
-
-            if axes_order == ['z', 'y', 'x']:
+            if axes_order == ["z", "y", "x"]:
                 self.array = np.swapaxes(array, 0, 2)
             else:
                 self.array = array
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -1,4 +1,5 @@
 from scipy.interpolate import RegularGridInterpolator
+import numpy as np
 from .profile import Profile
 
 
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -23,42 +23,33 @@ class FromArrayProfile(Profile):
     array : Array of the electric field of the laser pulse.
 
     axes : Python dictionary containing the axes vectors.
-        Keys are 'x', 'y', 't' if dim=xyt and 'r', 't' if dim=rt, respectively.
+        Keys are 'x', 'y', 't'.
         Values are the 1D arrays of each axis.
         array.shape = (axes['x'].size, axes['y'].size, axes['t'].size) in 3D,
         and similar in cylindrical geometry.
 
-    dim : Dimension of the data, 'xyt' or 'rt'
-
     axes_order : List of strings, giving the name and ordering of the axes in the array.
         Currently, only implemented for 3D, and supported values are
-        ['x', 'y', 'z'] and ['z', 'y', 'x'].
+        ['x', 'y', 't'] and ['t', 'y', 'x'].
     """
 
-    def __init__(self, wavelength, pol, array, axes, dim, axes_order=["x", "y", "z"]):
+    def __init__(self, wavelength, pol, array, axes, axes_order=["x", "y", "t"]):
         super().__init__(wavelength, pol)
 
-        assert dim == "xyt" or dim == "rt", "dim must be 'xyt' or 'rt'"
-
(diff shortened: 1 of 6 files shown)


MaxThevenet:
TODO:
- add test
- Get central wavelength from Hilbert transform

lasy/profiles/from_array_profile.py
```diff
@@ -0,0 +1,55 @@
...
+
+    array : Array of the electric field of the laser pulse.
+
+    axes : Python dictionary containing the axes vectors.
+        Keys are 'x', 'y', 't' if dim=xyt and 'r', 't' if dim=rt, respectively.
+        Values are the 1D arrays of each axis.
+        array.shape = (axes['x'].size, axes['y'].size, axes['t'].size) in 3D,
+        and similar in cylindrical geometry.
+
+    dim : Dimension of the data, 'xyt' or 'rt'
```
RemiLehe: Should we remove this argument, since the alternative is not supported?

lasy/profiles/from_array_profile.py
```diff
@@ -0,0 +1,55 @@
...
+    pol : list of 2 complex numbers (dimensionless)
+        Polarization vector. It corresponds to :math:`p_u` in the above
+        formula ; :math:`p_x` is the first element of the list and
+        :math:`p_y` is the second element of the list. Using complex
+        numbers enables elliptical polarizations.
+
+    array : Array of the electric field of the laser pulse.
+
+    axes : Python dictionary containing the axes vectors.
+        Keys are 'x', 'y', 't' if dim=xyt and 'r', 't' if dim=rt, respectively.
```
RemiLehe: Should we remove the mention of `r` and `t`, since this not currently supported?

lasy/profiles/from_array_profile.py
```diff
@@ -0,0 +1,55 @@
...
+
+    dim : Dimension of the data, 'xyt' or 'rt'
+    """
+
+    def __init__(self, wavelength, pol, array, axes, dim):
+        super().__init__(wavelength, pol)
+
+        assert dim == "xyt" or dim == "rt", "dim must be 'xyt' or 'rt'"
+
+        assert dim == "xyt", "Only dim='xyt' currently implemented"
```
RemiLehe: If we remove the `dim` argument, this would be removed too.
```suggestion
```

lasy/profiles/from_array_profile.py
```diff
@@ -0,0 +1,55 @@
...
+        defines :math:`\\omega_0` in the above formula, according to
+        :math:`\\omega_0 = 2\\pi c/\\lambda_0`.
+
+    pol : list of 2 complex numbers (dimensionless)
+        Polarization vector. It corresponds to :math:`p_u` in the above
+        formula ; :math:`p_x` is the first element of the list and
+        :math:`p_y` is the second element of the list. Using complex
+        numbers enables elliptical polarizations.
+
+    array : Array of the electric field of the laser pulse.
```
RemiLehe, line 23: My understanding from the code is that the first (i.e. slowest) axis of this array must correspond to `x`, and that the last (i.e. fastest) axis of this array must correspond to `t`. Is that correct? If it is the case, I think it is important to specify this in the docstring.

RemiLehe (changes requested):
Thanks for this PR!
I added a few inline comments. 
In addition, could you add a page for this new profile in the Sphinx documentation?

lasy/profiles/from_array_profile.py
MaxThevenet: This should be the symmetric to output, I'll take care of this.

lasy/profiles/from_array_profile.py
MaxThevenet: If that's fine with you, I suggest we keep it like this, and try to implement the rz version in the near future, so the API remains the same after this PR. But I can also remove this, and re-add when we implement the rz version.

MaxThevenet:
Thank you for your review! All points should be addressed now.

lasy/profiles/from_array_profile.py
MaxThevenet: In the end, I follow your suggestion and removed it.

lasy/profiles/from_array_profile.py
MaxThevenet: Both are now supported.

RemiLehe (approved):
Nice! Thanks for addressing the comments.

PR #144, opened 2023-05-22
Make wavelength optional for transverse profiles

RemiLehe:
This is a follow-up on https://github.com/LASY-org/lasy/pull/121, and allows user to omit the wavelength in the transverse profile. 
In addition, I also allow the user to set `z_foc` when calling `GaussianProfile`.


MaxThevenet (approved):
Great, thank!

Issue #145, opened 2023-05-22
Math formulas don't render properly in the documentation


RemiLehe:
When visiting this page:
https://lasydoc.readthedocs.io/en/latest/api/profiles/transverse/gaussian.html
I get this rendering:
<img width="312" alt="Screenshot 2023-05-22 at 10 01 27 AM" src="https://github.com/LASY-org/lasy/assets/6685781/29a3e2d7-4035-4be2-9885-6e5d892eb25e">
Is this specific to my browser/setup?

AngelFP:
I just had a look at the readthedocs history. Paradoxically, it seems that the problem was introduced in #120 😅. I'll try to find out the reason.

AngelFP:
Fixed in #146

PR #146, opened 2023-05-23
Fix issues in the documentation

AngelFP:
Fixes a few issues in the documentation, including the incorrect rendering of math expressions (#145), broken examples, and other minor problems.

To prevent these issues in the future, the documentation build will now fail if any warnings are raised.


MaxThevenet (approved):
Great, thanks for the cleaning! Let's see how this `fail_on_warning: true` works in practice.

MaxThevenet:
Could you resolve merge conflicts?

AngelFP:
Thanks for checking! It should be ready to merge now :)

PR #147, opened 2023-05-31
Close #142: Automatically normalize laser energy

### LINKED ISSUE #142: `laser_energy` not taken into account by default
It seems that, in the default `lasy` example, `laser_energy` is actually not taken into account.
This might be confusing to the user ; we should probably implement some default behavior, or a warning that tells the user when it is not used.

RemiLehe:


MaxThevenet (approved):
Thanks!

PR #139, opened 2023-05-18
Implement Gerchberg Saxton Algorithm

rob-shalloo:
This PR implements functionality to iteratively calculate laser phase given the laser amplitude in two planes and the distance between them. This is implemented through the Gerchberg-Saxton algorithm (https://en.wikipedia.org/wiki/Gerchberg%E2%80%93Saxton_algorithm) and address issue #59 

Contained within this pull request are 3 new files. 
- utils/zernike.py : a set of functions useful for calculating zernike polynomials on a 2D grid.
- utils/phase_retrieval.py : contains the function for performing the gerchberg saxton algorithm
- examples/example_gerchberg_saxton_algo.py an example showing the correct retrieval of the applied laser phase.

In the example file I:
- define a laser pulse at focus
- add some astigmatism to it
- propagate it forwards and backwards (deep-copying the laser object in each plane to get a new instance of the laser object)
- then feed these new laser objects on either side of focus into the `gerchberg_saxton_algo` function together with the distance between the planes
- the algorithm then proceeds and can be cut off either by a maximum number of itterations or by waiting until the summed residual of the field falls below a user-specified value.

The result is shown in the figure below

Creating this pull request has raised a few questions about how to properly implement a couple of features. I've made an attempt here, but perhaps a more thorough discussion is needed. These are:
- Extracting the laser field (for both amplitude and phase) is done by indexing into `laser.field.field`. This seems a little cumbersome. Should we add some functionality to extract amplitude, phase and intensity (for the purposes of calculation) or leave it all to open_PMD?
- It would be nice to be able to add (programmatically) different phase aberrations to a defined pulse, currently this is done as in the above point by indexing `laser.field.field` and redefining it or multiplying it by an `np.exp(1j*phase)`. Should we add functionality to manipulate the laser spatial phase in the same way one might manipulate the spectral phase to stretch the pulse?
- Currently I have to use two seperate instances of the laser object to represent the same laser pulse in two different planes. It would be incredibly useful if we could have everything in one object indexed by z position for example, although not sure how feasible this is.

Figure_1

PR #150, opened 2023-06-23
Merge Box and Grid into new class Field

MaxThevenet:
This PR proposes to merge classes `Box` and `Grid` into `Grid`, which contains both data and metadata for field. The list of member variables for `Grid` is exactly the union of that of previous `Box` and `Grid`, and its `__init__` function does the operations previously done in `Box.__init__` and `Grid.__init__`.

This terminology was agreed upon with @rob-shalloo @RemiLehe @hightower8083.

Finally, @rob-shalloo with this PR there is still 1 Field per Laser (Re: your first question in #139). Whether we put more is orthogonal to these changes, and will be discussed/done in a following PR.


rob-shalloo:
Hey Maxence, I think the merging of `Box` and `Grid` will certainly help to simplify the code and make it easier to add future modifications. 

I guess my first thought is can we think of a better name than `Field`. For example, in this structure the meaning of `laser.field.lo` and `laser.field.hi` are now confusing as they are not the minimum and maximum of the field amplitude for example, but the min and max spatial extent of the grid on which the field is based. Would it not make more sense to keep the name `Grid` and just merge the contents of `Box` into this. 

In this context, you break the laser into two key classes, the profile which still describes all the physics and the grid which describes the numerical realisation of the laser on a computational grid. The field then lives on the grid, so `laser.field.array` becomes `laser.grid.field` and the examples above become `laser.grid.lo` and `laser.grid.hi` which still make sense to me.

This is just an implementation suggestion though, I think the fundamental idea of combining these two classes is awesome.

MaxThevenet:
Alright, this is now ready for review again! If that's fine with everyone, let's try to merge it soon, so it does not conflict with too many PRs. Biggest conflict will be with #139 still, let's synchronise this.

RemiLehe (approved):
Looks good! 
Thanks a lot for these changes.

PR #151, opened 2023-07-02
Profile from openPMD file

MaxThevenet:
This PR creates a new profile to read a laser pulse from an openPMD file. It derives from `fromArrayProfile`. Works both when the profile read is a laser envelope or the electric field of a laser pulse. in the latter case, the envelope is obtained with a Hilbert transform. Both were tested and give the same result as shown in https://github.com/LASY-org/lasy/pull/141.

## Example
```python
from lasy.laser import Laser

# Create profile: one of
# - from FBPIC output
profile = FromOpenPMDProfile('lab_diags/hdf5/', 10, (0,1), 'E', 'y')
# - from LASY output
profile = FromOpenPMDProfile('./laserfbpic', 0, (1,0), 'laserEnvelope', envelope=True, prefix='laser3d')

# Create laser
dim = 'xyt'
lo = (-120e-6, -120e-6, -10e-15)
hi = (+120e-6, +120e-6, +60e-15)
npoints=(32,64,128)
laser = Laser(dim, lo, hi, npoints, profile)
laser.write_to_file('./laserfbpic/laser3d')

# Plot
from openpmd_viewer import OpenPMDTimeSeries
import numpy as np
ts = OpenPMDTimeSeries('./laserfbpic')
F, m = ts.get_field(field='laserEnvelope')
F = np.abs(F)
plt.figure(figsize=(8,8))
plt.imshow(F[:,:,F.shape[2]//2], aspect='auto')
plt.colorbar()
```
To do (probably in a separate PR):
 - [ ] Clean up handling of polarisation.
 - [ ] Some functions could be interesting in general, and could be in `tools` or (better IMO) openPMD-viewer.
 - [ ] Make compliant with latest version of the standard.
 - [ ] (only if needed) support all axes ordering.
 - [ ] The phase retrieval is 1D, which may cause problems for more complex laser pulses, so this could be improved.

### DIFF
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -80,13 +80,7 @@ class FromOpenPMDProfile(FromArrayProfile):
 
             # Get central wavelength from array
             phase = np.unwrap(np.angle(h))
-            omg0_h = (
-                -np.average(
-                    np.gradient(phase, axis=-1),
-                    weights=np.abs(h),
-                )
-                / dt
-            )
+            omg0_h = np.average(np.gradient(-phase, t, axis=-1), weights=np.abs(h) )
             wavelength = 2 * np.pi * c / omg0_h
             array = h * np.exp(1j * omg0_h * t)
         else:
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -43,16 +43,6 @@ class FromOpenPMDProfile(FromArrayProfile):
         Prefix of the openPMD file from which the envelope is read.
         Only used when envelope=True.
         The provided iteration is read from <path>/<prefix>_%T.h5.
-
-    wavelength : float (in meter)
-        The main laser wavelength :math:`\\lambda_0` of the laser, which defines
-        :math:`\\omega_0`, according to :math:`\\omega_0 = 2\\pi c/\\lambda_0`.
-        This argument is optional, and will overwrite the following default
-        behavior:
-        - if envelope is True, the central wavelength is read from the openPMD
-        file at openPMD envelope standard.
-        - if envelope is False, the central wavelength is obtained from the
-        Hilbert transform.
     """
 
     def __init__(
@@ -63,8 +53,7 @@ class FromOpenPMDProfile(FromArrayProfile):
         field,
         coord=None,
         envelope=False,
-        prefix=None,
-        wavelength=None,
+        prefix=None
     ):
         ts = OpenPMDTimeSeries(path)
         F, m = ts.get_field(iteration=iteration, field=field, coord=coord, theta=None)
@@ -105,23 +94,18 @@ class FromOpenPMDProfile(FromArrayProfile):
                 )
                 / dt
             )
-            wavelength_loc = 2 * np.pi * c / omg0_h
-            if wavelength is not None:
-                wavelength_loc = wavelength
+            wavelength = 2 * np.pi * c / omg0_h
             array = h * np.exp(1j * omg0_h * t)
         else:
-            if wavelength is None:
-                s = io.Series(path + "/" + prefix + "_%T.h5", io.Access.read_only)
-                it = s.iterations[iteration]
-                omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")
-                wavelength_loc = 2 * np.pi * c / omg0
-            else:
-                wavelength_loc = wavelength
+            s = io.Series(path + "/" + prefix + "_%T.h5", io.Access.read_only)
+            it = s.iterations[iteration]
+            omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")
+            wavelength = 2 * np.pi * c / omg0
             array = F
 
         axes_order = ["x", "y", "t"]
         super().__init__(
-            wavelength=wavelength_loc,
+            wavelength=wavelength,
             pol=pol,
             array=array,
             axes=axes,
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -1,0 +1,109 @@
+import numpy as np
+from scipy.signal import hilbert
+from scipy.constants import c
+from openpmd_viewer import OpenPMDTimeSeries
+from .profile import Profile
+from .from_array_profile import FromArrayProfile
+
+
+class FromOpenPMDProfile(Profile):
+    r"""
+    Profile defined from an openPMD file.
+
+    Parameters
+    ----------
+    path : string
+        Path to the openPMD file containing the laser field or envelope.
+        Passed directly OpenPMDTimeSeries.
+
+    iteration : int
+        Iteration at which the argument is read.
+        Passed directly OpenPMDTimeSeries.
+
+    pol : list of 2 complex numbers (dimensionless)
+        Polarization vector. It corresponds to :math:`p_u` in the above
+        formula ; :math:`p_x` is the first element of the list and
+        :math:`p_y` is the second element of the list. Using complex
+        numbers enables elliptical polarizations.
+
+    field : string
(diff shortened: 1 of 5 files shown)


MaxThevenet:
As suggested by @RemiLehe `FromOpenPMDProfile` now derives from `FromArrayProfile`.

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+        Name of the field containing the laser pulse
+        Passed directly OpenPMDTimeSeries.
+
+    envelope : boolean
+        Whether the file represents a laser envelope.
+        If not, the envelope is obtained from the electric field
+        using a Hilbert transform
+
+    prefix : string
+        Prefix of the openPMD file from which the envelope is red.
```
hightower8083: red -> read

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # If array does not contain the envelope but the electric field,
+        # extract the envelope with a Hilbert transform
+        if envelope == False:
+            # Assumes z is last dimension!
+            h = hilbert(F)
+
+            if "z" in m.axes.values():
+                # Flip to get complex envelope in t assuming z = -c*t
+                h = np.flip(h, axis=2)
```
hightower8083, line 77: should probably use converters

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+            {0: "x", 1: "y", 2: "t"},
+            {0: "t", 1: "y", 2: "x"},
+        ]
+
+        if m.axes in [{0: "z", 1: "y", 2: "x"}, {0: "t", 1: "y", 2: "x"}]:
+            F = F.swapaxes(0, 2)
+
+        if "z" in m.axes.values():
+            dt = m.dz / c
+            t = (m.z - m.z[0]) / c
```
hightower8083, line 64: here you implicitly set t[0]=0, while we usually center beam in time so `t_peak=0`

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+            if "z" in m.axes.values():
+                # Flip to get complex envelope in t assuming z = -c*t
+                h = np.flip(h, axis=2)
+
+            # Get central wavelength from array
+            phase = np.unwrap(np.angle(h))
+            omg0_h = (
+                -np.average(
+                    np.diff(phase),
+                    weights=0.5 * (np.abs(h[:, :, 1:]) + np.abs(h[:, :, :-1])),
```
hightower8083: here you could probably use `np.gradient(phase, axis=-1)` with `weights=np.abs(h)`

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+                )
+                / dt
+            )
+            wavelength_loc = 2 * np.pi * c / omg0_h
+            if wavelength is not None:
+                wavelength_loc = wavelength
+            array = h * np.exp(1j * omg0_h * t)
```
hightower8083: shouldn't we extract the `wavelength` in  case it is provided, i.e. overwrite `omg0_h  = 2 * np.pi * c /wavelength` ?

lasy/profiles/from_openpmd_profile.py
```diff
@@ -0,0 +1,129 @@
...
+        else:
+            if wavelength is None:
+                s = io.Series(path + "/" + prefix + "_%T.h5", io.Access.read_only)
+                it = s.iterations[iteration]
+                omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")
+                wavelength_loc = 2 * np.pi * c / omg0
+            else:
+                wavelength_loc = wavelength
+            array = F
```
hightower8083: maybe if wavelength  is provided and differs from the one in the field, we should physically  re-define the field with the new wavelength, e.g. 
```
            array = F
            s = io.Series(path + "/" + prefix + "_%T.h5", io.Access.read_only)
            it = s.iterations[iteration]
            omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")

            if wavelength is None:
                wavelength_loc = 2 * np.pi * c / omg0
            else:
                wavelength_loc = wavelength
                array *= np.exp(1j * (2 * np.pi * c  / wavelength - omg0) * t)
```

lasy/profiles/from_openpmd_profile.py
MaxThevenet: A typical workflow:
- Run a FBPIC simulation, get z output
- Read in LASY, and do this dummy conversion t=-z/c (needed as lasy requires t representation)
- Write envelope to file, t representation
- Read in from another code, and if needed do the inverse conversion (z = - c*t). Alternatively, this and the step above could be simplified if LASY could output z representation directly, see #155.

Points 2 and 4 are exactly symmetric, which is what we want, and that's easily done with this dummy convertor, so I think it should be like this. But the clean way would be to make this dummy convertor an option of the `import_from_z` convertor, that's for another PR (after #155 is addressed).

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Also to be discussed in #155.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Indeed

lasy/profiles/from_openpmd_profile.py
MaxThevenet: We should, isn't this already what the code does?

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Yes, we could do sth like this, I would keep this for another PR, as there are quite a few options that could make sense. I put it int #156.

MaxThevenet:
Thanks for your comments! I should have addressed them. I suggest that, for some of them, we address them in a separate PR, so I opened an issue (#156) accordingly. In particular, there should be CI, but that would be vastly simplified by either openPMD supporting t representation or by working a bit on the converter (also related to #155).

lasy/profiles/from_openpmd_profile.py
hightower8083: Well this: `array = h * np.exp(1j * omg0_h * t)`,  extracts from the field the frequency `omg0_h ` obtained from the Hilbert transform, which is a mean frequency for this particular spectrum. In a case if user wished to use some other central wavelength different from `2 * np.pi * c / omg0_h` (e.g. fix it 0.8um regardless to the actual spectrum) it will not do that -- the envelope will still be defined around the Hilbert one.

lasy/profiles/from_openpmd_profile.py
hightower8083: i'm not sure it's an separate option -- imagine, a user takes an envelope defined with some central frequency, but wants to obtain an object with a different one. In that case the envelope phase should be adjusted as suggested. Otherwise ` wavelength_loc = wavelength` is only passed to `super().__init__()` which registers it but never uses.

MaxThevenet:
Discussed offline: I'll remove the `wavelength` argument, as it is mostly here for debugging and for some corner cases, and implement some other utils (find central frequency, shift the frequency at which the envelope is defined, etc.) that we can call here.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: I deleted the wavelength argument

lasy/profiles/from_openpmd_profile.py
MaxThevenet: likewise, discussed offline, this option is useful and independent from this PR.

hightower8083 (approved):

hightower8083 (approved):

PR #157, opened 2023-08-02
Docs: Made clear-cut usage of laser.propagate

IlianCS:
I think using 'propagate' can lead to confusion as it controls the position of emission and not the direction of propagation.

### DIFF
--- a/docs/source/user_guide/index.rst
+++ b/docs/source/user_guide/index.rst
@@ -53,8 +53,7 @@ Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh 
    laser = Laser(dimensions,lo,hi,num_points,laser_profile)
 
 ..  code-block:: python
-   :caption: By default, the laser antenna will emit on the focal plan.
-   (Optional) Make the laser antenna emit outside the focal plan by one Rayleigh length.
+   :caption: By default, the laser antenna will emit on the focal plan. (Optional) Make the laser antenna emit outside the focal plan by one Rayleigh length.
 
    z_R            = 3.14159*spot_size**2/wavelength    # The Rayleigh length.
    laser.propagate(-z_R)                               # The laser antenna emits behind the focal plane
--- a/docs/source/user_guide/index.rst
+++ b/docs/source/user_guide/index.rst
@@ -21,7 +21,7 @@ First Example
 #############
 
 We will try a simple example to get familiar with the code structure and to verify the installation was successful.
-Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh length (the pulse is then located ahead of the focal plane) and then output it to a file.
+Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh length (the pulse is then located upstream of the focal plane) and then output it to a file.
 
 ..  code-block:: python
    :caption: First lets load in the required functions from the library.
@@ -56,7 +56,7 @@ Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh
    :caption: By default, the values of the laser envelope are injected on the focal plan. One can propagate it backwards by one Rayleigh length (optional).
 
    z_R            = 3.14159*spot_size**2/wavelength    # The Rayleigh length
-   laser.propagate(-z_R)                               # Propagate the pulse ahead of the focal plane
+   laser.propagate(-z_R)                               # Propagate the pulse upstream of the focal plane
 
 ..  code-block:: python
    :caption: Output the result to a file. Here we utilise the openPMD standard.
--- a/docs/source/user_guide/index.rst
+++ b/docs/source/user_guide/index.rst
@@ -53,10 +53,11 @@ Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh 
    laser = Laser(dimensions,lo,hi,num_points,laser_profile)
 
 ..  code-block:: python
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/index.rst
```diff
@@ -21,7 +21,7 @@ First Example
 #############
 
 We will try a simple example to get familiar with the code structure and to verify the installation was successful.
-Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh Length and then output it to file.
+Lets generate a Gaussian pulse at focus, emit it behind the focal plan by one Rayleigh length and then output it to a file.
```
MaxThevenet: Numerically, an array is allocated and initialised with the laser profile at focus. Then, using a spectral method, the laser is actually propagated backward, in 1 large time step `\Delta t = - z_R/c`. Therefore, it is my impression that the word propagate here is justified. Is the wording below better?
```suggestion
Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh length (the pulse is then located ahead of the focal plane) and then output it to a file.
```

docs/source/user_guide/index.rst
```diff
@@ -53,10 +53,10 @@ Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh
    laser = Laser(dimensions,lo,hi,num_points,laser_profile)
 
 ..  code-block:: python
-   :caption: Propagate the laser pulse backwards by one Rayeligh length.
+   :caption: By default, the laser antenna will emit on the focal plan. (Optional) Make the laser antenna emit outside the focal plan by one Rayleigh length.
```
MaxThevenet: LASY does not really have an antenna (defined as a plane onto which we generate currents so as to create the desired electromagnetic wave). IMO, we should therefore avoid this word here, so I would have a similar comment to the point above. But I agree with you, it would be nice to make this a bit more clear: I mentioned this above.

docs/source/user_guide/index.rst
MaxThevenet: For clarification, one could add a sentence here, explaining that the output file can then be used as an input for a simulation tool, what do you think?

docs/source/user_guide/index.rst
```diff
@@ -53,10 +53,10 @@ Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh
...
 
    z_R            = 3.14159*spot_size**2/wavelength    # The Rayleigh length.
-   laser.propagate(-z_R)
+   laser.propagate(-z_R)                               # The laser antenna emits behind the focal plane
```
MaxThevenet: Likewise, this step actually does a numerical propagation by 1 large time step, and this is not use an antenna per se, so I think this wording could be confusing.

MaxThevenet (commented):
Thanks for this PR, some help in the documentation certainly cannot hurt! However, I believe the word "propagate" is more accurate than "antenna" in some places, I tried to clarify this with more comments below. I hope this is a bit more clear, do you think you could modify your PR to incorporate the changes (and thanks for the typos fixes)?

docs/source/user_guide/index.rst
IlianCS: It is already mentionned that: "This file may now be viewed, copied, shared or used as an input to a variety of other simulation tools." (at the end of the page)

docs/source/user_guide/index.rst
IlianCS: "By default, the values of the laser envelope are injected on the focal plan" does that sound better?

IlianCS:
Thanks a lot for your review, @MaxThevenet.
I added your suggestions and completed with some remarks, please let me know if some other changes have to be made.

docs/source/user_guide/index.rst
```diff
@@ -21,7 +21,7 @@ First Example
 #############
 
 We will try a simple example to get familiar with the code structure and to verify the installation was successful.
-Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh Length and then output it to file.
+Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh length (the pulse is then located ahead of the focal plane) and then output it to a file.
```
MaxThevenet:
```suggestion
Let's generate a Gaussian pulse at focus, propagate it backwards by one Rayleigh length (the pulse is then located upstream of the focal plane) and then output it to a file.
```

docs/source/user_guide/index.rst
```diff
@@ -53,17 +53,17 @@ Lets generate a Gaussian pulse at focus, propagate it backwards by one Rayeligh
    laser = Laser(dimensions,lo,hi,num_points,laser_profile)
 
 ..  code-block:: python
-   :caption: Propagate the laser pulse backwards by one Rayeligh length.
+   :caption: By default, the values of the laser envelope are injected on the focal plan. One can propagate it backwards by one Rayleigh length (optional).
 
-   z_R            = 3.14159*spot_size**2/wavelength    # The Rayleigh length.
-   laser.propagate(-z_R)
+   z_R            = 3.14159*spot_size**2/wavelength    # The Rayleigh length
+   laser.propagate(-z_R)                               # Propagate the pulse ahead of the focal plane
```
MaxThevenet:
```suggestion
   laser.propagate(-z_R)                               # Propagate the pulse upstream of the focal plane
```

MaxThevenet (approved):
Looks good, thanks!

MaxThevenet (approved):

Issue #127, opened 2023-04-24
Milestone: from FBPIC output to HiPACE++ input


MaxThevenet:
This needs the following operations:
1. Generate a RZ profile from the output of an FBPIC simulation
2. Convert from RZ to 3D
3. Write 3D output at lasy format
4. Read profile in HiPACE++

This requires the following new capabilities:
- [ ] (required, in lasy) Read input from openPMD file. It is not clear to me whether we should have 1 reader per code or 1 for all openPMD-based files. I would recommend trying the latter, and we could pass a few arguments or just the code name to fix the parameters that may differ between codes (z or t representation, field stored, E vs. a0, RZ vs. 3D, etc.)
- [ ] (required, in lasy) dummy converter z = -c*t. As HiPACE++ currently reverts this operation exactly, that could be a trick to make the this z-to-z pipeline fully accurate and computationally cheap. [Maxence]
- [ ] (required, in lasy, but can rely on openPMD-viewer) Convert from electric field to electric field envelope. [Maxence]
- [ ] Convert from RZ to 3D (this is probably already implemented)

There are other ways to make this work, including supporting RZ input in HiPACE++, but I think it is preferable to do such 
operations in LASY at the moment.

PR #158, opened 2023-08-03
Improvements in the reader from openPMD file

MaxThevenet:
This PR proposes the following improvements, all meant to achieve #127:
- The option to output the normalized vector potential instead of the electric field is implemented. For the rationale, see https://github.com/openPMD/openPMD-standard/pull/281. This is certainly open for discussion. Converters from E to a0 and reverse are implemented in utils (it is a very vanilla version, definitely room for improvement there).
- The calculation of local and average frequency is in a separate function. This way, we can implement different methods to do this operation (e.g. FFT-based) easily.
- profile from_array and from_openPMD now also support RZ representation. Without this, operating on the 3D data was not practical. Again, very vanilla implementation, the implementation assumes axial symmetry for the envelope, the mode decomposition is not included in this PR.

With these changes, the output of an FBPIC simulation can be sent as the input of a HiPACE++ simulation. The figure below shows, for a laser pulse in a parabolic channel, the pulse width during propagation, when starting the HiPACE++ simulation from the output of the FBPIC simulation at different times. Significant differences can be observed, which could very much come from differences between the two codes: even without using LASY the two codes give slightly different results for this case. This is probably a matter of tuning the resolution, box size, time step etc., so I do not think it is necessarily a problem.
laser_width_hipace_fbpic
[laser_width_hipace_fbpic.pdf](https://github.com/LASY-org/lasy/files/12252598/laser_width_hipace_fbpic.pdf)

The scripts I used are posted below, so we can later work on the agreement. I had to add some .txt to upload the files.

[inputs_fbpic.py.txt](https://github.com/LASY-org/lasy/files/12252686/inputs_fbpic.py.txt)
[inputs_hipace.txt](https://github.com/LASY-org/lasy/files/12252687/inputs_hipace.txt)
[script_fbpic.sh.txt](https://github.com/LASY-org/lasy/files/12252689/script_fbpic.sh.txt)
[script_hipace.sh.txt](https://github.com/LASY-org/lasy/files/12252690/script_hipace.sh.txt)

The lasy handling was done using:
```python
from lasy.laser import Laser
from openpmd_viewer.addons import LpaDiagnostics
import scipy.constants as scc

profile2 = FromOpenPMDProfile('./DATA', 26, (0,1), 'E', 'y', theta=0)

dim = 'xyt'
lo = (-100e-6, -100e-6, 0)
hi = (+100e-6, +100e-6, (m0.zmax-m0.zmin)/scc.c)
npoints=(128, 128, 500)
## Tested, also works with
# dim = 'rt'
# lo = (0, 0)
# hi = (+100e-6, (m0.zmax-m0.zmin)/scc.c)
# npoints=(300, 600)
laser = Laser(dim, lo, hi, npoints, profile2)
laser.write_to_file('./lasy_t3/laser3d', use_a0=True)
```

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -8,7 +8,7 @@ from lasy.utils.grid import Grid
 from lasy.utils.laser_utils import (
     normalize_energy,
     normalize_peak_field_amplitude,
-    normalize_peak_intensity,
+    normalize_peak_intensity
 )
 from lasy.utils.openpmd_output import write_to_openpmd_file
 
@@ -398,7 +398,7 @@ class Laser:
             self.grid.field = prop.z2t(transform_data, t_axis, z0=z0, t0=t0).T
             self.grid.field *= np.exp(1j * (z0 / c + t_axis) * omega0)
 
-    def write_to_file(self, file_prefix="laser", file_format="h5"):
+    def write_to_file(self, file_prefix="laser", file_format="h5", use_a0=False):
         """
         Write the laser profile + metadata to file.
 
@@ -417,4 +417,5 @@ class Laser:
             self.grid,
             self.profile.lambda0,
             self.profile.pol,
+            use_a0
         )
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -398,7 +398,9 @@ class Laser:
             self.grid.field = prop.z2t(transform_data, t_axis, z0=z0, t0=t0).T
             self.grid.field *= np.exp(1j * (z0 / c + t_axis) * omega0)
 
-    def write_to_file(self, file_prefix="laser", file_format="h5", save_as_vector_potential=False):
+    def write_to_file(
+        self, file_prefix="laser", file_format="h5", save_as_vector_potential=False
+    ):
         """
         Write the laser profile + metadata to file.
 
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -398,7 +398,7 @@ class Laser:
             self.grid.field = prop.z2t(transform_data, t_axis, z0=z0, t0=t0).T
             self.grid.field *= np.exp(1j * (z0 / c + t_axis) * omega0)
 
-    def write_to_file(self, file_prefix="laser", file_format="h5", use_a0=False):
+    def write_to_file(self, file_prefix="laser", file_format="h5", save_as_vector_potential=False):
         """
         Write the laser profile + metadata to file.
 
@@ -409,6 +409,10 @@ class Laser:
 
         file_format : string
             Format to be used for the output file. Options are ``"h5"`` and ``"bp"``.
+
+        save_as_vector_potential : bool (optional)
+            Whether the envelope is converted to normalized vector potential
+            before writing to file.
         """
         write_to_openpmd_file(
             self.dim,
@@ -417,5 +421,5 @@ class Laser:
             self.grid,
             self.profile.lambda0,
             self.profile.pol,
-            use_a0,
+            save_as_vector_potential,
         )
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -8,7 +8,7 @@ from lasy.utils.grid import Grid
 from lasy.utils.laser_utils import (
     normalize_energy,
     normalize_peak_field_amplitude,
-    normalize_peak_intensity
+    normalize_peak_intensity,
 )
 from lasy.utils.openpmd_output import write_to_openpmd_file
 
@@ -417,5 +417,5 @@ class Laser:
             self.grid,
             self.profile.lambda0,
             self.profile.pol,
-            use_a0
+            use_a0,
         )
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -5,6 +5,7 @@ import openpmd_api as io
 from openpmd_viewer import OpenPMDTimeSeries
 from .from_array_profile import FromArrayProfile
 from lasy.utils.laser_utils import get_frequency
+from lasy.utils.grid import Grid
 
 
 class FromOpenPMDProfile(FromArrayProfile):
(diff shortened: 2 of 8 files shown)


MaxThevenet:
Tests pass, this is ready for review :)

lasy/profiles/from_openpmd_profile.py
```diff
@@ -43,56 +45,171 @@ class FromOpenPMDProfile(FromArrayProfile):
...
+                omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")
+                wavelength = 2 * np.pi * c / omg0
+                array = F
+
+            axes_order = ["r", "t"]
+
+        print(
+            "Wavelength used in the definition of the envelope (nm):",
+            wavelength * 1.0e9,
+        )
```
AngelFP: Could we avoid this print by default? maybe with a `verbose` parameter.

lasy/utils/laser_utils.py
```diff
@@ -205,3 +207,157 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+    else:
+        assert dim in ["xyt", "rt"]
+        if dim == "xyt" and not phase_unwrap_1d:
+            print("WARNING: using 3D phase unwrapping, this can be expensive")
+
+        if not is_hilbert:
+            h = np.squeeze(hilbert(grid.field))
+        else:
+            h = np.squeeze(grid.field)
+        print(h.shape)
```
AngelFP: Probably a leftover print from debugging

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Thanks! This should be fixed

AngelFP:
Results from a similar test restarting an FBPIC simulation with Wake-T. Looking promising so far!

image

MaxThevenet:
Thanks @AngelFP. The laser propagation is quite well-captured. Final iteration below. Unfortunately, the agreement still isn't great, but nothing that can obviously be attributed to LASY, probably some convergence question.
fig

lasy/profiles/from_openpmd_profile.py
```diff
@@ -43,56 +45,176 @@ class FromOpenPMDProfile(FromArrayProfile):
...
-        else:
-            t = m.t
-        axes = {"x": m.x, "y": m.y, "t": t}
-
-        # If array does not contain the envelope but the electric field,
-        # extract the envelope with a Hilbert transform
-        if envelope == False:
-            # Assumes z is last dimension!
-            h = hilbert(F)
+        F, m = ts.get_field(iteration=iteration, field=field, coord=coord, theta=theta)
```
MaxThevenet, line 81: In principle, we could avoid depending on the viewer here, but that would take a bit more work (in particular to properly convert from RZ decomposition of fields to RZ decomposition of envelope). I would be happy to discuss this.

lasy/laser.py
```diff
@@ -398,7 +398,7 @@ def import_from_z(self, field_z, z_axis, z0=0.0, t0=0.0, backend="NP"):
             self.grid.field = prop.z2t(transform_data, t_axis, z0=z0, t0=t0).T
             self.grid.field *= np.exp(1j * (z0 / c + t_axis) * omega0)
 
-    def write_to_file(self, file_prefix="laser", file_format="h5"):
+    def write_to_file(self, file_prefix="laser", file_format="h5", use_a0=False):
         """
         Write the laser profile + metadata to file.
```
AngelFP, line 406: Could you add this to the docstring here too
```
use_a0 : bool (optional)
    Whether the envelope is converted to normalized vector potential
    before writing to file.
```

Also, perhaps `a0` can be a bit confusing (i.e., as the maximum of a), maybe something like `use_a` or more explicit like `save_as_potential`.

lasy/profiles/from_openpmd_profile.py
```diff
@@ -43,56 +45,176 @@ class FromOpenPMDProfile(FromArrayProfile):
         Prefix of the openPMD file from which the envelope is read.
         Only used when envelope=True.
         The provided iteration is read from <path>/<prefix>_%T.h5.
+
```
AngelFP, line 48: There's one vertical space too much here

lasy/profiles/from_openpmd_profile.py
```diff
@@ -43,56 +45,176 @@ class FromOpenPMDProfile(FromArrayProfile):
...
+            else:
+                t = m.t
+            axes = {"x": m.x, "y": m.y, "t": t}
+
+            if phase_unwrap_1d is None:
+                phase_unwrap_1d = True
+
+            # If array does not contain the envelope but the electric field,
+            # extract the envelope with a Hilbert transform
+            if envelope == False:
```
AngelFP:
```suggestion
            if not envelope:
```

lasy/utils/laser_utils.py
```diff
@@ -205,3 +207,165 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+
+    # Filter out too small frequencies
+    omega = np.maximum(omega, lower_bound * central_omega)
+    # Filter out too large frequencies
+    omega = np.minimum(omega, upper_bound * central_omega)
+
+    return omega, central_omega
+
+
+def field_to_a0(grid, omega0):
```
AngelFP: Same comment as above. Maybe `a0` can be misleading. `field_to_potential` could also work, for example. Not a big issue in any case.

AngelFP (approved):

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Thanks! I fixed it.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: fixed

MaxThevenet:
Thanks @AngelFP for your review, I addressed your comments in the last commits.

AngelFP:
Great, thanks!

PR #159, opened 2023-08-07
Publish to PyPI on release

AngelFP:
Adds a workflow that will take care of automatically uploading `lasy` to PyPI every time we make a new release.

The workflow uses the new Trusted Publishers method from PyPI (https://docs.pypi.org/trusted-publishers/), which does not require any username/password or token.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,35 @@
+name: Publish package to PyPI
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build-n-publish:
+    name: Build and publish package to PyPI
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
```
MaxThevenet, line 19: Is there a reason for Python 3.8 specifically?

MaxThevenet (approved):
Thanks! LGTM, see minor question below.

.github/workflows/publish-to-pypi.yml
AngelFP: Not particularly. It is recommended to specify a python version (https://github.com/actions/setup-python), but any version lasy supports would be fine.

Issue #155, opened 2023-07-31
More features to the z-t converters


MaxThevenet:
`Laser.export_to_z` and `Laser.import_from_z` do proper conversions between Z and T representations. Some additions would make it useful:
- Include the I/O handling so LASY can output Z and T representations
- Include the I/O handling so LASY can input Z and T representations
- Allow dummy conversion (z = -ct) and decide on where the 0 is
- Maybe this could be put to utils? That would lighten the `Laser` class, but might be cumbersome, not sure.

hightower8083:
Concerning the last point, I would move propagate and converters (all `axiprop`-dependent stuff) to an utility class somewhere separate (e.g. `propagate_util.py`) and is inherited by lasy. This can also be joined with #164, as in case of non-installed `axiprop`, this class is replaced with a dummy, so these methods will simply not appear in the lasy object and we can avoid these multiple `assert axiprop_installed` conditions.

MaxThevenet:
I think whether we do a dummy z-t conversion or a full clean conversion using axiprop should be independent on whether axiprop is installed. For FBPIC -> LASY -> Wake-T workflow, we will always do the dummy conversion as it is more accurate. So the options I see are:
1. put all in `laser_utils.py`, and keep the `assert axiprop_installed` somewhere. If that could be manageable, I would prefer it this way, so that we always call `convert_z_to_t` (or whatever name), as it is the same operation, and pass an argument whether it is dummy.
2. just have 2 functions: `convert_z_to_t` and `convert_z_to_t_dummy` (and same for `t_to_z`).

I would have a preference for the first one, as it should be slightly easier to use and compare both, but either works.

PR #164, opened 2023-08-14
Make `axiprop` optional

AngelFP:
This PR proposes making `axiprop` an optional dependency, since it is only needed for the laser propagation features. This would make `lasy` a lighter dependency for other libraries or codes that only need the basic lasy features (e.g., laser profile generation, openPMD readers, laser analysis, etc.).

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -4,6 +4,7 @@ from scipy.constants import c
 try:
     from axiprop.lib import PropagatorFFT2, PropagatorResampling
     from axiprop.containers import ScalarFieldEnvelope
+
     axiprop_installed = True
 except ImportError:
     axiprop_installed = False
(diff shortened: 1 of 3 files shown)


hightower8083:
Well `axiprop` only relies on `numpy` and `scipy`, that we also have in `lasy`, and it only takes ~200K on the disk, so it isn't much of a heavy dependency =)
Anyway, if needed, I'd suggest we do this along with some restructuring clean-up in laser class as discussed in #155

AngelFP:
True, I thought it was actually heavier. I would still try to keep it to a minimum, but I'm happy to discuss this in the next meeting. As an example, I'm currently implementing a laser analysis module in VisualPIC that relies on lasy, but which does not need any of the axiprop features.

An alternative implementation to the assertion checks would be to define something like a `@requires_axiprop` decorator that can be applied to any function that depends on axiprop.

MaxThevenet:
From the discussion on Monday: If there's an easy way with `pip` to install a package _removing_ an optional dependency (there is the option to add an optional dependency, but not clear there's a way to remove one), then we'll make Axiprop optional. Otherwise, for the moment, we'll keep it as a required dependency, as it is lightweight.

PR #166, opened 2023-08-16
Add new utils and reimplement `FromOpenPMDProfile`

AngelFP:
This PR adds a few handy utilities to process arrays from openPMD (actually, from the openPMD-viewer) and a new `field_to_envelope` method. With these new utilities, the implementation of `FromOpenPMDProfile` can be greatly simplified.

In addition, the `envelope` parameter of `FromOpenPMDProfile` has been removed. The class itself will determine whether the input field is an envelope by checking whether the array is complex.

### DIFF
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -36,11 +36,6 @@ class FromOpenPMDProfile(FromArrayProfile):
         Name of the field containing the laser pulse
         Passed directly OpenPMDTimeSeries.
 
-    envelope : boolean
-        Whether the file represents a laser envelope.
-        If not, the envelope is obtained from the electric field
-        using a Hilbert transform
-
     prefix : string
         Prefix of the openPMD file from which the envelope is read.
         Only used when envelope=True.
@@ -71,7 +66,6 @@ class FromOpenPMDProfile(FromArrayProfile):
         pol,
         field,
         coord=None,
-        envelope=False,
         prefix=None,
         theta=None,
         phase_unwrap_1d=None,
@@ -96,7 +90,7 @@ class FromOpenPMDProfile(FromArrayProfile):
 
         # If array does not contain the envelope but the electric field,
         # extract the envelope with a Hilbert transform
-        if not envelope:
+        if not np.iscomplexobj(F):
             grid = create_grid(F, axes, dim)
             grid, omg0 = field_to_envelope(grid, dim, phase_unwrap_1d)
             array = grid.field[0]
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -85,7 +85,7 @@ class FromOpenPMDProfile(FromArrayProfile):
             if phase_unwrap_1d is None:
                 phase_unwrap_1d = False
             axes_order = ["r", "t"]
-        
+
         F, axes = reorder_array(F, m, dim)
 
         # If array does not contain the envelope but the electric field,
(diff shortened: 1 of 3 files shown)


lasy/profiles/from_openpmd_profile.py
```diff
@@ -81,128 +74,32 @@ def __init__(
...
-                wavelength = 2 * np.pi * c / omg0
-                array = F
-
             axes_order = ["r", "t"]
 
+        F, axes = reorder_array(F, m, dim)
+
+        # If array does not contain the envelope but the electric field,
+        # extract the envelope with a Hilbert transform
+        if not np.iscomplexobj(F):
```
MaxThevenet: This assumes that `is_complex iff is_envelope`. This could be dangerous, and maybe a safer option would be to check if `envelopeField` exists and has the right value, as discussed [in the openPMD-standard PR](https://github.com/openPMD/openPMD-standard/blob/aeadcda9d71ff702d7cb412b04595f76b0d42105/EXT_LaserEnvelope.md)? Until this is in the standard and the codes have adopted it,  we could add the optional argument, which would overwrite the test `envelopeField` test. What do you think?

lasy/utils/laser_utils.py
```diff
@@ -242,17 +241,12 @@ def get_frequency(
     is_envelope : bool (optional)
         Whether the field provided uses the envelope representation, as used
         internally in lasy. If False, field is assumed to represent the
-        electric field.
+        Hilbert transform of the electric field.
```
MaxThevenet: Is there a reason for changing this behavior? I can explain the rationale for the original behavior: this function (and others in laser_utils) aims at being relatively general and easy-to-use: the only lasy-specific constraint is to create a `Grid` object. Asking the users to know about the Hilbert transform may be raising the barrier a bit high. For this reason, the function was able to take care of the Hilbert transform itself, and the option to pass the Hilbert transform was a way to avoid re-computing it multiple times.

lasy/utils/openpmd_input.py
```diff
@@ -0,0 +1,149 @@
+import numpy as np
+import scipy.constants as ct
+
+from .grid import Grid
+
+
+def reorder_array(array, md, dim):
```
MaxThevenet, line 5: These 3 functions do 2 things:
- swap axes so the last axis is time.
- convert (in a dummy way) from z representation to t representation.

The second operation (conversion) is not really about reordering. However, in a future PR I will probably move the conversion to a different place (put z-t converters in utils, and give them the option to do dummy). Therefore, I am fine with the changes you propose, this is just FYI.

lasy/utils/openpmd_input.py
```diff
@@ -0,0 +1,149 @@
...
+    axes = {"r": r, "t": t}
+
+    array = 0.5 * (
+        array[array.shape[0] // 2 :, :]
+        + np.flip(array[: array.shape[0] // 2, :], axis=0)
+    )
+    return array, axes
+
+
+def create_grid(array, axes, dim):
```
MaxThevenet: Could this go to `laser_utils.py`? The function is not specific to openPMD format, as it only takes basic numpy and python types, which is very handy. So I believe it could be a great helper for users to apply the `laser_utils` functions to their data, regardless of the context.

MaxThevenet (changes requested):
Great, thanks for this PR! See comments below. Happy to discuss.

AngelFP:
Thanks for the review @MaxThevenet. The last commits should address all comments.

MaxThevenet (approved):
Looks great, thanks!

PR #168, opened 2023-08-18
Add new `get_spectrum` and `get_duration` utils

AngelFP:
As the title says, this PR adds two new laser analysis methods. They currently support grids that contain an envelope. In addition to them, a new convenience method to get the grid cell volume has also been implemented.

### DIFF
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -44,7 +44,7 @@ def compute_laser_energy(dim, grid):
 
     if dim == "xyt":
         energy = ((dV * epsilon_0 * 0.5) * abs(envelope) ** 2).sum()
-    elif dim == "rt":
+    else:  # dim == "rt":
         energy = (
             dV[np.newaxis, :, np.newaxis]
             * epsilon_0
@@ -263,7 +263,7 @@ def get_spectrum(grid, dim, bins=20, range=None, omega0=None, phase_unwrap_1d=No
     dV = get_grid_cell_volume(grid, dim)
     if dim == "xyt":
         weights = np.abs(grid.field) * dV
-    elif dim == "rt":
+    else:  # dim == "rt":
         weights = np.abs(grid.field) * dV[np.newaxis, :, np.newaxis]
     # Get weighted spectrum.
     # Neglects the 2 first and last time slices, whose values seems to be
@@ -399,7 +399,7 @@ def get_duration(grid, dim):
     dV = get_grid_cell_volume(grid, dim)
     if dim == "xyt":
         weights = np.abs(grid.field) * dV
-    elif dim == "rt":
+    else:  # dim == "rt":
         weights = np.abs(grid.field) * dV[np.newaxis, :, np.newaxis]
     # project weights to longitudinal axes
     weights = np.sum(weights, axis=(0, 1))
@@ -533,7 +533,7 @@ def get_grid_cell_volume(grid, dim):
     dz = grid.dx[-1] * c
     if dim == "xyt":
         dV = grid.dx[0] * grid.dx[1] * dz
-    elif dim == "rt":
+    else:  # dim == "rt":
         r = grid.axes[0]
         dr = grid.dx[0]
         # 1D array that computes the volume of radial cells
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -213,7 +213,7 @@ def get_spectrum(
     bins=20,
     is_envelope=True,
     omega0=None,
-    on_axis=False,
+    mode='sum'
 ):
     """
     Get the the frequency spectrum of an envelope or electric field.
@@ -251,14 +251,15 @@ def get_spectrum(
         Angular frequency at which the envelope is defined. Required if
         `is_envelope=True`.
 
-    on_axis : bool (optional)
-        Whether to get the spectrum on axis or, otherwise, the summed spectrum
-        along the transverse direction(s).
+    mode : {'sum', 'on_axis', 'full'} (optional)
+        Whether to return the summed spectrum along the transverse direction(s),
+        the on-axis spectrum, or the full (unsumed) spectrum in the whole
+        transverse domain.
 
     Returns
     -------
     spectrum : ndarray of doubles
-        Array with the spectrum amplitudes (FFT amplitude * dt).
+        Array with the spectral density in J/(rad/s).
 
     omega_spectrum : scalar
         Array with the frequencies of the spectrum.
@@ -268,7 +269,7 @@ def get_spectrum(
     omega_spectrum = 2 * np.pi * freq
 
     # Get on axis or full field.
-    if on_axis:
+    if mode == 'on_axis':
         if dim == "xyt":
             nx, ny, nt = grid.field.shape
             field = grid.field[nx // 2, ny // 2]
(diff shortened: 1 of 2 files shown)


lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+    """
+    Get the the frequency spectrum of an envelope.
+
+    Parameters
+    ----------
+    grid : a Grid object.
+        It contains a ndarrays with the field data from which the
+        spectrum is computed, and the associated metadata. The last axis must
+        be the longitudinal dimension.
+        Can be the full electric field or the envelope.
```
MaxThevenet: Can it really be either? The `get_frequency` call below contains `is_envelope=True`.

MaxThevenet: Whether this supports both envelope and full field is something we could discuss. I believe both would be nice.

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
     return env, ext
 
 
+def get_spectrum(grid, dim, bins=20, range=None, omega0=None, phase_unwrap_1d=None):
```
MaxThevenet: Just a comment here: the spectrum could also be obtained from a Fourier transform, which would work very easily with field or envelope (in the latter case, one would just need to add `omega0` to the spectrum) and could be more accurate, so it will be very interesting to compare both. I am sure @hightower8083 would share this view :)

But this is definitely not required in this PR, can be added later.

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+    bins : int (optional)
+        Number of bins of the spectrum.
+
+    range : list of float (optional)
+        List of two values indicating the minimum and maximum frequency of the
+        spectrum.
+
+    omega0 : scalar
+        Angular frequency at which the envelope is defined.
+        Required if an only if is_envelope is True.
```
MaxThevenet: likewise

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+
+    phase_unwrap_1d : boolean (optional)
+        Whether the phase unwrapping is done in 1D.
+        This is not recommended, as the unwrapping will not be accurate,
+        but it might be the only practical solution when dim is 'xyt'.
+
+    Returns
+    -------
+    spectrum : ndarray of doubles
+        Array with the spectrum amplitudes.
```
MaxThevenet: Is the spectrum calibrated, or just normalized? Calibrated would be really great, so if it's the case could you mention the units here?

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+
+    omega_spectrum : scalar
+        Array with the frequencies of the spectrum.
+    """
+    # Get frequency array.
```
MaxThevenet: Maybe it's good to specify it at least once, we are slightly inaccurate in the wording.
```suggestion
    # Get the array of angular frequency.
```

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,78 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+    )
+    # Calculate weights of each frequency (amplitude of the field).
+    dV = get_grid_cell_volume(grid, dim)
+    if dim == "xyt":
+        weights = np.abs(grid.field) * dV
+    else:  # dim == "rt":
+        weights = np.abs(grid.field) * dV[np.newaxis, :, np.newaxis]
+    # Get weighted spectrum.
+    # Neglects the 2 first and last time slices, whose values seems to be
+    # slightly off (maybe due to lower order derivative at the edges).
```
```suggestion
    # Neglects the 2 first and last time slices, whose values seem to be
    # slightly off (maybe due to lower-order derivative at the edges).
```

MaxThevenet (commented):
Looks great, thanks for this PR! See comments below.

lasy/utils/laser_utils.py
```diff
@@ -313,6 +380,32 @@ def get_frequency(
...
+        The grid with the envelope to analyze.
+    dim : str
+        Dimensionality of the grid.
+
+    Returns
+    -------
+    float
+        RMS duration of the envelope in seconds.
+    """
+    # Calculate weights of each frequency (amplitude of the field).
```
AngelFP:
```suggestion
    # Calculate weights of each grid cell (amplitude of the field).
```

lasy/utils/laser_utils.py
AngelFP: I agree. It would be nice to support several methods here as well as in other utils (e.g., `get_duration`` could have RMS, FWHM, fit, etc. options).

lasy/utils/laser_utils.py
AngelFP: I copy-pasted this docstring too fast. Currently it supports only envelope. I think we should make a general decision about this, so that all methods support both or only one representation.

AngelFP:
New approach using spectral density (squaring the FFT amplitude to get J/(rad/s)).
on axis:
image

summed:
image

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,145 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
     return env, ext
 
 
+def get_spectrum(
+    grid, dim, range=None, bins=20, is_envelope=True, omega0=None, method="sum"
+):
+    r"""
+    Get the the frequency spectrum of an envelope or electric field.
```
```suggestion
    Get the frequency spectrum of an envelope or electric field.
```

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,145 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+        - 'rt' : The laser pulse is represented on a 2D grid:
+                 Cylindrical (r) transversely, and temporal (t) longitudinally.
+
+    range : list of float (optional)
+        List of two values indicating the minimum and maximum frequency of the
+        spectrum. If provided, only the FFT spectrum within this range
+        will be returned using interpolation.
+
+    bins : int (optional)
+        Number of bins of to which to interpolate the spectrum if a `range`
```
MaxThevenet: dunno
```suggestion
        Number of bins into which to interpolate the spectrum if a `range`
```

lasy/utils/laser_utils.py
```diff
@@ -211,6 +206,145 @@ def get_full_field(laser, theta=0, slice=0, slice_axis="x", Nt=None):
...
+        is given.
+
+    is_envelope : bool (optional)
+        Whether the field provided uses the envelope representation, as used
+        internally in lasy. If False, field is assumed to represent the
+        the electric field.
```
```suggestion
        the full electric field (with fast oscillations).
```

MaxThevenet (commented):
Looks great, thanks for the tests and extensive docstrings! See minor points below.

MaxThevenet (approved):

PR #169, opened 2023-08-18
Update docs to use `sphinx_design`

AngelFP:
This PR updates the grid we show in the home page of the documentation. It was using a package (`sphinx_panels`) that is now deprecated in favor of `sphinx_design`. This fixes an issue where the dark color scheme was not properly working.

Before:
image

Now:
image


MaxThevenet (approved):
Thanks for this PR!

PR #171, opened 2023-08-21
Conform to the new openPMD standard

RemiLehe:
This makes the lasy openPMD file conform with https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_LaserEnvelope.md


ax3l (approved):

MaxThevenet (approved):
Thanks for this PR!

PR #174, opened 2023-08-23
Make scikit-image an optional dependency, and change default for phase unwrapping

MaxThevenet:
Quick tests by @AngelFP show that 1D phase unwrapping can be sufficient. This PR proposes to keep the option of ND phase unwrapping, but turn it off by default and hence make scikit-image an optional dependency.

### DIFF
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -2,8 +2,10 @@ import numpy as np
 from scipy.constants import c, epsilon_0, e, m_e
 from scipy.interpolate import interp1d
 from scipy.signal import hilbert
+
 try:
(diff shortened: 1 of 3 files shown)


lasy/utils/laser_utils.py
```diff
@@ -2,10 +2,15 @@
 from scipy.constants import c, epsilon_0, e, m_e
 from scipy.interpolate import interp1d
 from scipy.signal import hilbert
-from skimage.restoration import unwrap_phase
+
+try:
```
ax3l: You can move this import closer to the usage, which will make imports of lasy faster (if scikit-image is present but not used)

ax3l (approved):
Thank you, looks great 👍

PR #177, opened 2023-08-24
Do not extrapolate field in `FromArrayProfile` when using `rz` geometry

AngelFP:
Currently, in order to avoid issues in the first cell when using `rz` geometry, the array is extrapolated when a point outside of the
original domain is requested. However, this leads to artifacts at the upper boundary (and potentially also in the time boundaries).

This PR fixes this issue by creating an interpolation array that is mirrored along the axis. This allows the interpolation within the first cell to be correct and let's us use a fill value of 0 outside of the other boundaries.


MaxThevenet (approved):
Looks good, thanks for the fix!

PR #179, opened 2023-09-07
Update writing of `position`

RemiLehe:
The previous syntax for writing `position` was ambiguous about the type of `position` and could result in unreadable types when the `lasy` file is produced on certain architectures.

The new syntax now uses the default numpy type (`float64`).

### DIFF
--- a/lasy/utils/openpmd_output.py
+++ b/lasy/utils/openpmd_output.py
@@ -103,7 +103,7 @@ def write_to_openpmd_file(
     # Define the dataset
     dataset = io.Dataset(data.dtype, data.shape)
     env = m[io.Mesh_Record_Component.SCALAR]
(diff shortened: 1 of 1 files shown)


lasy/utils/openpmd_output.py
```diff
@@ -103,7 +103,7 @@ def write_to_openpmd_file(
     # Define the dataset
     dataset = io.Dataset(data.dtype, data.shape)
     env = m[io.Mesh_Record_Component.SCALAR]
-    env.position = [0] * len(dim)
+    env.position = np.zeros(len(dim))
```
ax3l:
```suggestion
    env.position = np.zeros(len(dim), dtype=np.float64)
```

ax3l (approved):
Thank you! :)

PR #180, opened 2023-09-07
Add info on lasy version

RemiLehe:
This adds the software version and allows other codes (e.g. WarpX, HiPACE++, FBPIC, Wake-T) to check for the right lasy version.


ax3l (approved):
Thank you! :)

PR #182, opened 2023-09-13
Space time profile

spencerjolly:
This adds the ability to create a pulse with space-time couplings. Spatial chirp at the focus (z=0) can be added with a real-valued parameter 'sc'. This uses the highest level 'Profile' class. An interesting test is if the propagator works properly to propagate such pulses to the beginning of a simulation box. Analytical equations could also be used to directly calculate the fields at a user-defined z-position.

### DIFF
--- a/lasy/profiles/spacetime_profile.py
+++ b/lasy/profiles/spacetime_profile.py
@@ -90,17 +90,15 @@ class SpaceTimeProfile(Profile):
         """
         transverse = np.exp(-(x**2 + y**2) / self.w0**2)
 
-        tau_eff = np.sqrt(self.tau**2 + (2 * self.sc / self.w0)**2)
+        tau_eff = np.sqrt(self.tau**2 + (2 * self.sc / self.w0) ** 2)
 
         spacetime = np.exp(
-        	-(t - self.t_peak
-        	+ (2 * 1j * self.sc * x / self.w0**2))**2 / tau_eff**2
-        	)
+            -((t - self.t_peak + (2 * 1j * self.sc * x / self.w0**2)) ** 2)
+            / tau_eff**2
+        )
 
-        oscillatory = *np.exp(
-        	1.0j * (self.cep_phase - self.omega0 * (t - self.t_peak))
-        	)
+        oscillatory = *np.exp(1.0j * (self.cep_phase - self.omega0 * (t - self.t_peak)))
 
-        envelope = transverse*spacetime*oscillatory
+        envelope = transverse * spacetime * oscillatory
 
         return envelope
--- a/lasy/profiles/spacetime_profile.py
+++ b/lasy/profiles/spacetime_profile.py
@@ -9,56 +9,56 @@ class SpaceTimeProfile(Profile):
 
     More precisely, the electric field corresponds to:
 
-        .. math::
+    .. math::
 
-            E_u(\\boldsymbol{x}_\\perp,t) = Re\\left[ E_0\\,
-            \\exp\\left(-\\frac{\\boldsymbol{x}_\\perp^2}{w_0^2}
-            - \\frac{(t-t_{peak}+2ibx/w_0^2)^2}{\\tau_{eff}^2}
-            - i\\omega_0(t-t_{peak}) + i\\phi_{cep}\\right) \\times p_u \\right]
+        E_u(\\boldsymbol{x}_\\perp,t) = Re\\left[ E_0\\,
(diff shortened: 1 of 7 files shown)


lasy/profiles/spacetime_profile.py
```diff
@@ -0,0 +1,102 @@
+import numpy as np
+
+from .profile import Profile
+
+
+class SpaceTimeProfile(Profile):
```
RemiLehe, line 6: Maybe this could actually be part of the `GaussianProfile` laser. See other comments.

lasy/profiles/spacetime_profile.py
```diff
@@ -0,0 +1,102 @@
...
+        formula. Note that :math:`\\tau = \\tau_{FWHM}/\\sqrt{2\\log(2)}`,
+        where :math:`\\tau_{FWHM}` is the Full-Width-Half-Maximum duration
+        of the intensity distribution of the pulse.
+
+    w0: float (in meter)
+        The waist of the laser pulse, i.e. :math:`w_0` in the above formula.
+
+    sc: spatial chirp, b in the above formula, that results in a mixing
+        of the longitudinal and transverse profiles. Must be in units
+        of [x/omega]. A representative real value is b = w0 * tau.
```
RemiLehe: Here is a suggestion that is more consistent with the other docstrings:
```suggestion
    b: float (in meter.second)
        Spatial chirp, i.e. :math:`b` in the above formula, that results in the laser frequency
        varying as a function of `x`. A representative real value is b = w0 * tau.

```

RemiLehe (changes requested):
Thanks for this PR!

I added a few inline comments. In addition, I have a few high-level comments:
- Would it be OK to include the parameter `b` in the [`GaussianProfile`](https://lasydoc.readthedocs.io/en/latest/api/profiles/gaussian.html) and have its value be 0 by default, instead of having a separate `SpaceTimeProfile` class? (This is the approach that we adopted for the built-it Gaussian profile in WarpX: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Laser/LaserProfilesImpl/LaserProfileGaussian.cpp#L99 ; it would require reimplementing the `GaussianProfile` differently in WarpX.)
- Could you add an automated test, e.g. [here](https://github.com/LASY-org/lasy/blob/development/tests/test_laser_profiles.py)? The test could for instance check for instance that the laser duration is unchanged in the presence of spatial chirp, or (even better) check that the laser frequency is indeed `x` dependent.

lasy/profiles/spacetime_profile.py
spencerjolly: I don't actually see how it could be applied with GaussianProfile. In that class it calls the temporal and spatial profiles separately, but we need access to the space and time dimensions at the same time.

RemiLehe: Yes, agreed. If we decide to have it applied with GaussianProfile, we would completely change the implementation of the `GaussianProfile` (which is definitely possible): instead of calling the temporal and spatial profile separately, we would have the Python code for the gaussian profile directly in the body of the `GaussianProfile` class. Does that work for you?

spencerjolly: Yes, ok with me. In fact, it would mostly entail pasting the code from this PR to the GaussianProfile class. However, I also wanted to add a pulse-front tilt parameter (in addition to the current "spatial chirp" parameter). I could do that in this PR so that it is all there, and then we could paste it.

Let me know how you want to continue.

RemiLehe: Sounds good. I think it is good to do all of the above as part of the current PR, i.e. adding spatial chirp, and moving all of this to the `GaussianProfile` class.

spencerjolly: I've done these changes. There are some weird test and docs errors at the moment, but in any case I need to build a stand-alone test for the space-time cases. I will try to do it next week.

spencerjolly:
@RemiLehe I am slowly progressing on this. Right now I am finishing the tests for angular dispersion and spatial chirp. How can I select the field only at a given transverse position? Basically I want to test either the central frequency or t_peak given spatial chirp or angular dispersion, and for that I would like to have the temporal envelop at y=0 and x=w0.

MaxThevenet:
Thanks for this PR! I think PR #307 implements these changes. @spencerjolly could you check if this works for you?

spencerjolly:
@MaxThevenet seems good to me. Really nice to see that it could be implemented together with the tests.

PR #193, opened 2023-10-18
Show import path in the documentation

RemiLehe:
When reading the documentation for various profiles (e.g. `SuperGaussianTransverseProfile`), the users will also need to know how to import these profiles in their `lasy` script.

This PR makes sure that the Sphinx documentation does show the import path. I also used the shorter path that is available for a number of classes (e.g. `lasy.profiles.transverse.SuperGaussianTransverseProfile` instead of `lasy.profiles.transverse.super_gaussian_profile.SuperGaussianTransverseProfile`

This is how the documentation looks in the **development** branch:

Screenshot 2023-10-18 at 4 00 09 PM

This is how it looks **in this PR**:

Screenshot 2023-10-18 at 4 00 33 PM

### DIFF
--- a/docs/source/api/profiles/combined_profile.rst
+++ b/docs/source/api/profiles/combined_profile.rst
@@ -1,5 +1,5 @@
 Combined Longitudinal and Transverse Profile
 ============================================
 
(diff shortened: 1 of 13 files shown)


AngelFP (approved):
Looks good! Thanks for taking care of using the shorter paths.

PR #194, opened 2023-10-18
Do not show low-level objects in documentation

RemiLehe:
Most users will probably read the API section of the documentation to understand how the use the different laser profiles.

At the same time, the API section of the documentation also shows some low-level function (base classes for the profiles, various utility functions that we only use internally in `lasy` but are not supposed to be used directly by the user). The few advanced users that would want to interact with these functions would probably look them up directly on Github, instead of using the Sphinx documentation for this.

This PR removes some of the documentation for the low-level utility functions, and moves some others to the `code_overview` section.


docs/source/api/index.rst
```diff
@@ -9,6 +9,5 @@ The sections below describe the main objects that are accessible in the ``lasy``
 
    laser
    profiles/index
-   utils/index
```
RemiLehe: This removes low-level utilities from the doc.

docs/source/api/profiles/index.rst
```diff
@@ -1,12 +1,6 @@
 Laser Profiles
 ==============
 
-Laser pulses are incorporated into *lasy* via the ``Profile`` class.
-Typically a laser will be constructed through a combination of two classes ``LongitudinalProfile`` and ``TransverseProfile`` which represent the longitudinal and transverse profiles  of the laser.
-
-.. autoclass:: lasy.profiles.profile.Profile
-    :members:
-
```
RemiLehe, line 9: Removes base class (not used by the user directly)

docs/source/api/profiles/longitudinal/index.rst
```diff
@@ -1,9 +1,6 @@
 Longitudinal Profiles
 =====================
 
-.. autoclass:: lasy.profiles.longitudinal.longitudinal_profile.LongitudinalProfile
-    :members:
-
```
RemiLehe, line 6: Removes base class

docs/source/api/profiles/transverse/index.rst
```diff
@@ -1,9 +1,6 @@
 Transverse Laser Profiles
 =========================
 
-.. autoclass:: lasy.profiles.transverse.transverse_profile.TransverseProfile
-    :members:
-
```
RemiLehe, line 6: Removes base class

AngelFP:
Looks like the docs are failing because the `code_overview/grid.rst` and `code_overview/openpmd_output.rst` files are not included in any toctree. Could you see if they can be included in one?

RemiLehe:
Thanks for reviewing this PR!
In the end, I opted to simply remove the documentation for the `Grid` object. (I think that it was added in the documentation at a time when we did require the user to initialize a `Grid` object explicitly, but this is no longer the case.)

AngelFP:
Thanks for the changes. With respect to the grid, I think it might be useful to keep it (although we could take care of this in a separate PR later on) because it is the base class with which all the laser diagnostics tools in https://github.com/LASY-org/lasy/blob/development/lasy/utils/laser_utils.py operate, so the user still needs to create one to use them. What do you think @MaxThevenet ?

MaxThevenet:
Looks great, thanks for the cleaning! I agree with Ángel, the user can use all functions in utils without creating a `Laser` object, all they need is to initialise a `Grid` object. Therefore, it would be nice to keep just this one in the documentation.

RemiLehe:
OK, sounds good. I added `Grid` back in the documentation (but moved it out of the `utils` section of the documentation, since it was the only remaining file in this section ; if we later add more files we could recreate this section.)

AngelFP:
Thanks Remi, can you show were the grid is documented now? I can't find it in the updated docs.

RemiLehe:
Ah, sorry about this, it seems that I accidentally removed it when struggling with the ReadTheDocs CI.

In the end, I added back in the `utils` folder, and I also added documentation for the Gerchberg-Saxton algorithm.

AngelFP (approved):
Looks good now, thanks!

PR #197, opened 2023-10-20
Include examples as notebooks in the documentation

RemiLehe:
This PR uses `nbsphinx` so that we can use notebooks as source files in the Sphinx documentation.

The notebooks are then nicely formatted in the chosen style of the documentation. In addition, they are run by `readthedocs` as part of building the documentation - which provides one form of automatic testing. (The readthedocs build will fail if the notebook is obsolete and does not work anymore.)

In particular, our first example (gaussian laser) has now been converted in notebook format.

Future PRs will add more notebooks in the `tutorials` folder.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,6 +3,7 @@
 -e .
 matplotlib
 numpydoc
+jupyter
 nbsphinx
 pydata-sphinx-theme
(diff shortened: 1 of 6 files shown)


AngelFP (approved):
Looks good, thanks for this!

PR #199, opened 2023-10-25
Add thin optical elements in `lasy`

RemiLehe:
This PR is one suggested way to add thin optical elements in `lasy`, through which the laser profiles can be propagated. This is heavily inspired by examples in `axiprop`.

Thin elements are represented by their complex multiplying factor `T(x, y, omega)` and are therefore applied after performing one Fourier transform on the laser profile, along the time axis. This PR only implements one optical element (`ParabolicMirror`), but more will be added next.

~~For this reason, I chose to have the user pass the optical element to the `propagate` function (since this function starts by performing a Fourier transform along the time axis)~~ In the latest version, the API has been changed.


AngelFP:
This is great. I know it's early work, but would it make sense to be able to define something like an `OpticalBeamline` that can have several `OpticalElements`? Then we could pass this beamline to `propagate`.

lasy/laser.py
```diff
@@ -160,6 +160,46 @@ def normalize(self, value, kind="energy"):
...
+
+        Parameter
+        ---------
+        optical_element: an :class:`.OpticalElement` object (optional)
+            Represents a thin optical element, through which the laser
+            propagates.
+        """
+        # Transform the field from temporal to frequency domain
+        time_axis_indx = -1
+        field_fft = np.fft.ifft(self.grid.field, axis=time_axis_indx, norm="backward")
```
RemiLehe, line 175: The implementation is not efficient currently, as it performs FFTs forward and backward.
This will be fixed in #256, by keeping the Fourier-space fields in memory and having a flag that keeps track of this.

lasy/laser.py
```diff
@@ -160,6 +160,46 @@ def normalize(self, value, kind="energy"):
         else:
             raise ValueError(f'kind "{kind}" not recognized')
 
+    def apply_optics(self, optical_element):
```
RemiLehe, line 163: New API

lasy/optical_elements/optical_element.py
```diff
@@ -0,0 +1,41 @@
...
+
+        .. math::
+
+            \tilde{\mathcal{E}}_{out}(x, y, \omega) = T(x, y, \omega)\tilde{\mathcal{E}}_{in}(x, y, \omega)
+
+        Parameters
+        ----------
+        x, y, omega: ndarrays of floats
+            Define points on which to evaluate the multiplier.
+            These arrays need to all have the same shape.
```
MaxThevenet, line 31: Should we consider 1D arrays?

lasy/optical_elements/optical_element.py
```diff
@@ -0,0 +1,41 @@
...
+
+        Returns
+        -------
+        multiplier: ndarray of complex numbers
+            Contains the value of the multiplier at the specified points
+            This array has the same shape as the arrays x, y, omega
+        """
+        # The base class only defines dummy multiplier
+        # (This should be replaced by any class that inherits from this one.)
+        return np.zeros_like(x, dtype="complex128")
```
MaxThevenet: Open question, that would mean the base class does nothing to the laser.
```suggestion
        return np.ones_like(x, dtype="complex128")
```

lasy/laser.py
```diff
@@ -160,6 +160,46 @@ def normalize(self, value, kind="energy"):
...
+
+        # Create the frequency axis
+        dt = self.grid.dx[time_axis_indx]
+        omega0 = self.profile.omega0
+        Nt = self.grid.field.shape[time_axis_indx]
+        omega = 2 * np.pi * np.fft.fftfreq(Nt, dt) + omega0
+
+        # Apply optical element
+        if self.dim == "rt":
+            r, w = np.meshgrid(self.grid.axes[0], omega, indexing="ij")
```
MaxThevenet: Minor comment: might be a bit dangerous to mix `w` and `omega`.

lasy/laser.py
```diff
@@ -160,6 +160,46 @@ def normalize(self, value, kind="energy"):
...
+            # The line below assumes that amplitude_multiplier
+            # is cylindrically-symmetric, hence we pass
```
MaxThevenet:
```suggestion
            # is cylindrically symmetric, hence we pass
```

lasy/laser.py
MaxThevenet: (same as comment in `optical_element.py`: should the `meshgrid` be done in the optical element? Could just be an option)

lasy/laser.py
```diff
@@ -169,10 +209,16 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
         distance : scalar
             Distance by which the laser pulse should be propagated
 
+        initial_optical_element: an :class:`.OpticalElement` object (optional)
+            Represents a thin optical element, through which the laser
+            propagates, before propagating for `distance` in free space.
+            If this is `None`, no optical element is used.
+
```
MaxThevenet: Should this be removed?
```suggestion
```

MaxThevenet (commented):
Thanks for this PR!

lasy/optical_elements/optical_element.py
RemiLehe: That's a good question.
I think it is easier for people to add new optical elements if `x`, `y`, `omega` all have the same shape. (In that case, people can simply write the formula for the multiplier using element-wise operations, without having to broadcast.) I think that we adopted a similar strategy for the laser profiles.

lasy/optical_elements/optical_element.py
RemiLehe: Ok, yes, good point.

lasy/laser.py
RemiLehe: Ah, yes, sorry. I thought I had removed it.

lasy/laser.py
```diff
@@ -160,6 +160,51 @@ def normalize(self, value, kind="energy"):
...
+            r, omega = np.meshgrid(self.grid.axes[0], omega_1d, indexing="ij")
+            # The line below assumes that amplitude_multiplier
+            # is cylindrically symmetric, hence we pass
+            # `r` as `x` and 0 as `y`
+            multiplier = optical_element.amplitude_multiplier(r, 0, omega)
+            # The azimuthal modes are the components of the Fourier transform
+            # along theta (FT_theta). Because the multiplier is assumed to be
+            # cylindrically symmetric (i.e. theta-independent):
+            # FT_theta[ multiplier * field ] = multiplier * FT_theta[ field ]
+            # Thus, we can simply multiply each azimuthal mode by the multiplier.
```
RemiLehe, line 194: I added more explanations here.

MaxThevenet (approved):

PR #202, opened 2023-10-26
Resampling

NadezhdaKHACHAT:
Adds the option to resample the field onto a new grid during the propagation step. This can be useful for example when focusing the laser from a large beam size to a small spot.

The resampling option only works in the `'rt'` geometry. In order to resample the fields, the user can simply pass a new grid object to the `propagate()` function. The new grid replaces the original grid.

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -292,24 +292,13 @@ class Laser:
         self.grid.field *= np.exp(1j * omega0 * distance / c)
 
     def write_to_file(
-        self,
-        iteration,
-        distance,
-        file_prefix="laser",
-        file_format="h5",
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks a lot for this PR! This looks very useful.
Could you add an automated test of this feature, for instance by adding one more test function in this file:
https://github.com/LASY-org/lasy/blob/development/tests/test_gaussian_propagator.py

RemiLehe:
@hightower8083 mentioned that it would be good to first merge https://github.com/LASY-org/lasy/pull/227

RemiLehe:
Superseeded by #269

Issue #183, opened 2023-10-08
Plotting laser profiles


berceanu:
Is it a good idea to also include plotting functionality, for being able to inspect the various laser profiles?

PR #203, opened 2023-10-26
Add `show` method to visualize the laser

RemiLehe:
This addresses #183. This adds a basic plotting capability (useful e.g. to make sure that the laser is fully inside the box, before dumping to disk). 
For more advanced plotting, the user should probably dump the laser to file, and then use openPMD-viewer.

I also made use of this method in the tutorial:
Screenshot 2023-10-26 at 11 10 29 AM

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -279,9 +279,9 @@ class Laser:
 
     def show(self, **kw):
         """
-        Show a 2D image of the laser amplitude
+        Show a 2D image of the laser amplitude.
 
-        Parameters:
+        Parameters
         ----------
         **kw: additional arguments to be passed to matplotlib's imshow command
         """
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -277,8 +277,7 @@ class Laser:
             save_as_vector_potential,
         )
 
-
-    def show( self, **kw ):
+    def show(self, **kw):
         """
         Show a 2D image of the laser amplitude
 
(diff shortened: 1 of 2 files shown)


lasy/laser.py
```diff
@@ -276,3 +276,46 @@ def write_to_file(
...
+        """
+        if self.dim == "rt":
+            # Show field above and below axis, with proper sign
+            E = [
+                np.concatenate(
+                    ((-1) ** m * self.grid.field[0, ::-1], self.grid.field[0])
+                )
+                for m in self.grid.azimuthal_modes
+            ]
+            E = sum(E)
```
```suggestion
            E = sum(E) # Sum all the modes
```

lasy/laser.py
```diff
@@ -276,3 +276,46 @@ def write_to_file(
...
+    def show(self, **kw):
+        """
+        Show a 2D image of the laser amplitude.
+
+        Parameters
+        ----------
+        **kw: additional arguments to be passed to matplotlib's imshow command
+        """
+        if self.dim == "rt":
+            # Show field above and below axis, with proper sign
```
```suggestion
            # Show field in the plane y=0, above and below axis, with proper sign for each mode
```

hightower8083:
do you think we could have here some kind of `envelope=True` option and in the case of `False` show the carrier-frequency resolved field from `get_full_field`?

AngelFP (approved):
Hi @RemiLehe, thanks for this.

Overall it looks good to me, and it makes me think again that we should make the `Grid` a smarter class (e.g., so that it knows its dimensions, extent, etc.). It is otherwise quite cumbersome to work with it.

MaxThevenet:
@hightower8083 I would be in favor of only supporting the envelope in general, and avoid the carrier frequency, as it does not bring more information but is much more expensive. It's not a strong opinion, would be happy to hear everyone's thoughts.

RemiLehe:
Yes, we can discuss this in the next meeting, and, if needed, do a follow-up PR. For now, since this is approved, I will merge.

Issue #213, opened 2023-12-09
XYT Laser Propagate with Nx != Ny


ax3l:
Hi,

I tried to resolve a laser profile for `xyt` differently in X than in Y and realized this breaks the propagator with:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[52], line 2
      1 profile_focal_distance = 27.0e-6 
----> 2 laser.propagate(-profile_focal_distance)  # Propagate the pulse upstream of the focal plane

File ~/micromamba/envs/lasy-warpx/lib/python3.12/site-packages/lasy/laser.py:237, in Laser.propagate(self, distance, nr_boundary, backend, show_progress)
    235     # Propagate the spectral image
    236     transform_data = np.transpose(field_fft).copy()
--> 237     self.prop.step(
    238         transform_data, distance, overwrite=True, show_progress=show_progress
    239     )
    240     field_fft[:, :, :] = np.transpose(transform_data).copy()
    242 # Choose the time translation assuming propagation at v=c

File ~/micromamba/envs/lasy-warpx/lib/python3.12/site-packages/axiprop/steppers.py:65, in StepperNonParaxial.step(self, u, dz, overwrite, show_progress)
     63 phase_loc = self.kz[ikz]**2 - self.kr2
     64 phase_loc = self.bcknd.sqrt( (phase_loc>=0.)*phase_loc )
---> 65 self.u_ht *= self.bcknd.exp( 1j * dz * phase_loc )
     67 self.iTST()
     68 u_step[ikz] = self.bcknd.to_host(self.u_iht)

ValueError: operands could not be broadcast together with shapes (50,150) (150,50) (50,150) 
```

Code snippet for this:
```py
spot_size = 5e-6
pulse_duration = 30e-6

dimensions     = 'xyt'                               # Use 3D geometry
lo             = (-5*spot_size, -5*spot_size, -7*pulse_duration)           # Lower bounds of the simulation box
hi             = ( 5*spot_size,  5*spot_size, 10*pulse_duration)  # Upper bounds of the simulation box
num_points     = (150, 50, 250)                     # Number of points in each dimension

laser = Laser(dimensions, lo, hi, num_points, laser_profile)

profile_focal_distance = 27.0e-6 
laser.propagate(-profile_focal_distance)  # Propagate the pulse upstream of the focal plane
```

hightower8083:
[...]

PR #214, opened 2023-12-10
fix the bug with array shapes in xyt propagator

hightower8083:
This fixes the bug #213 with array shapes in xyt propagator -- `np.transpose` is replaced with `np.moveaxis`.

- [x] fix `xyt`
- [x] double-check `rt`


hightower8083:
I think this one is good to go:
- propagator and converters for `xyt` are fixed 
- tests are now considering `Nx!=Ny`  for all these methods
- no problem with `rt`: initial (lasy)  `[m, r, t]` representation is processed in a loop `for m: transpose([r, fft[t]])` which gives proper `[ω, r]` modes shapes to be passed to axiprop. Transposition here is applied to 2D `[r, ω]` shapes and cannot miss

MaxThevenet (approved):
Thanks for this PR!

PR #216, opened 2023-12-11
Closes #215 , add overloading for addition and multiplication operators in TransverseProfile

### LINKED ISSUE #215: Overloading Addition and Multiplication Operators of TransverseProfile
Currently in the definition of the laser `profile` the addition and multiplication operators have been overloaded thanks to some additions by Sören. However, we should also consider adding this functionality to the `transverse_profile` as for the decomposition/reconstruction of the laser profile into Hermite Gaussians, only the transverse profile is required and currently performing these calculations is quite slow and cumbersome with the full `profile` object.

rob-shalloo:
I've followed @soerenjalas implementation of overloading in the Profile class and added it directly in the TransverseProfile class. This allows one to create a laser transverse profile from a sum of hermite gaussians without having to first create the full laser object. This makes things cleaner and should hopefully speed up calculations.

This should close #215

### DIFF
--- a/lasy/profiles/transverse/__init__.py
+++ b/lasy/profiles/transverse/__init__.py
@@ -4,7 +4,12 @@ from .laguerre_gaussian_profile import LaguerreGaussianTransverseProfile
 from .super_gaussian_profile import SuperGaussianTransverseProfile
 from .jinc_profile import JincTransverseProfile
 from .transverse_profile_from_data import TransverseProfileFromData
-from .transverse_profile import TransverseProfile, SummedTransverseProfile, ScaledTransverseProfile
+from .transverse_profile import (
+    TransverseProfile,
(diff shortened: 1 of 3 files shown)


soerenjalas:
I think it would make sense to add some basic testing for these function. It would be good to have at least something like [this](https://github.com/LASY-org/lasy/blob/0cf9cac56f16ecac2a0aae9dd04d12c520594b1a/tests/test_laser_profiles.py#L238).

rob-shalloo:
Hey @soerenjalas , thanks for the great suggestion. I've added the tests now, again shamelessly copying your code... but seems to work ;)

soerenjalas (approved):

MaxThevenet (approved):
Great, thanks for this PR!

PR #217, opened 2023-12-12
remove the version specifier for axiprop in requirements

hightower8083:
This PR removes version specifier for `axiprop` in requirements, since no API breaking developments in there expected, and the releases are checked for compatibility with `lasy` anyway


ax3l (approved):
LGTM, thank you! :+1:

PR #227, opened 2024-02-06
Create, register axiprop containers and use them in propagator

hightower8083:
Adding axiprop container to the laser object to maintain consistently the fields in temporal and frequency domains

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -2,12 +2,14 @@ import numpy as np
 from scipy.constants import c
 
 from axiprop.lib import PropagatorFFT2, PropagatorResampling
+from axiprop.containers import ScalarFieldEnvelope
 
 from lasy.utils.grid import Grid
 from lasy.utils.laser_utils import (
     normalize_energy,
     normalize_peak_field_amplitude,
     normalize_peak_intensity,
+    get_container,
 )
 from lasy.utils.openpmd_output import write_to_openpmd_file
 
@@ -139,7 +141,8 @@ class Laser:
         else:
             raise ValueError(f'kind "{kind}" not recognized')
 
-    def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True):
+
+    def propagate(self, distance, nr_boundary=0, backend="NP", show_progress=True):
         """
         Propagate the laser pulse by the distance specified.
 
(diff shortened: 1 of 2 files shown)


hightower8083:
@MaxThevenet, this should be ready for review/merge

MaxThevenet (commented):
Thanks for this PR! A few general comments:
- "container" is a generic term for this kind of libraries. Could we keep a more specific name? `FieldsAxprp` or so, similar to what it was before, would do.
- In 3D, `container` is a `ScalarFieldEnvelope`, whereas in RZ+ it is a list of `ScalarFieldEnvelope`s, so a different type. Do we have a standardised approach to this point? In `Grid`, the array just has 1 more dimension, which is not applicable here. It could be a list in both cases for instance.

lasy/utils/laser_utils.py
```diff
@@ -682,6 +682,59 @@ def create_grid(array, axes, dim):
...
+
+    backend : string (optional)
+        Backend used by axiprop (see AVAILABLE_BACKENDS in axiprop
+        documentation for more information).
+    """
+    time_axis_indx = -1
+
+    if dim == "rt":
+        # Make a list of containers for RT geometry
+        Container = []
```
MaxThevenet, line 720: I think we capitalise e.g. class names, not lists here, so should this really be capitalised?

lasy/laser.py
```diff
@@ -152,47 +155,29 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
...
-                ]
-                self.grid.field[:, -nr_boundary:, :] *= absorb_layer_shape[
-                    None, :, None
-                ]
-                self.grid.field[:, :nr_boundary, :] *= absorb_layer_shape[::-1][
-                    None, :, None
-                ]
-
-        # Transform the field from temporal to frequency domain
-        field_fft = np.fft.ifft(self.grid.field, axis=time_axis_indx, norm="backward")
```
RemiLehe, line 187: Where is this operation being done, in the new code?

RemiLehe: Ah, I see, it is in `get_container` now.

lasy/utils/laser_utils.py
```diff
@@ -682,6 +682,59 @@ def create_grid(array, axes, dim):
     return grid
 
 
+def get_container(dim, grid, omega0, transform=True, n_dump=0, backend="NP"):
+    """
+    Export field to axiprop container, clean, and trasnform to frequency.
```
RemiLehe:
```suggestion
    Export field to axiprop container, clean, and transform to frequency.
```

RemiLehe:
Thanks for this PR.

Do I understand correctly that the main aim of this PR is to avoid calling the temporal-to-spectral space FFT multiple times, in cases where we call `progagate` multiple times? (which should result in faster computation)

lasy/laser.py
RemiLehe: I wonder if we should make this more explicit in the code.

RemiLehe:
Maybe a few questions to discuss:
- [ ] The data in `container` now duplicates the data in `fields`. Should we consider keeping only the container (as a follow-up PR)?
- [ ] The general way in which we are caching container (and propagator) objects in between calls to `propagate` could be a bit error prone (e.g. users/developers could modify `fields` without modify the corresponding `container` object). Or some users might not be aware that the option `force_sync` is available. I am wondering if there is a better way.

RemiLehe:
Replaced by https://github.com/LASY-org/lasy/pull/256

PR #230, opened 2024-02-13
Fix complex interpolation in FromArrayProfile method

hightower8083:
This PR fixes the issue with using scipy's `RegularGridInterpolator` with complex numbers used in `FromArrayProfile`. While the amplitude and phase of the complex field normally vary smoothly in space and time, the real and imaginary parts typically strongly oscillate, which results in erroneous results. 

See here an example of the field read from measurement file: on the original grid (left) and with time axis refined by 4
image

This PR replaces the single interpolation operation with real-valued amplitude and phase (unwrapped along `t`) interpolations. This fixes the issue with refined grid:
image


MaxThevenet (approved):
Looks good, thanks!

PR #239, opened 2024-03-29
A fix to prevent division by zero during normalization

hightower8083:
I have recently added `import_from_lasy` and `export_to_lasy` methods in axiprop and for the latter one I have used an empty profile to get it's field rewritten, i.e.:
```python
wavelength = 2 * np.pi * c / Container.omega0
empty_longitudinal_profile = LongitudinalProfile(wavelength=wavelength)
empty_transverse_profile = TransverseProfile()
empty_profile = CombinedLongitudinalTransverseProfile(
    wavelength=wavelength,
    pol=polarization,
    laser_energy=0.0,
    long_profile=empty_longitudinal_profile,
    trans_profile=empty_transverse_profile,
)
```

With this I've noticed that the normalization forced at initialization of `CombinedLongitudinalTransverseProfile` divides field by `0.0`,  and generates the corresponding warning. Although my methods work fine with this (`None` field values are overwritten anyway), I don't like this warning. 

So this PR checks the energy before normalization and skips it if field is zero.

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -8,6 +8,7 @@ from lasy.utils.laser_utils import (
     normalize_energy,
     normalize_peak_field_amplitude,
     normalize_peak_intensity,
+    compute_laser_energy,
 )
 from lasy.utils.openpmd_output import write_to_openpmd_file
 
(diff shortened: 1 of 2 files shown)


ax3l (approved):
LGTM - thanks a lot!

MaxThevenet (commented):
IIRC, we (in particular @RemiLehe) discussed offline and suggested that the check could be in `Laser.normalize`, or maybe in each `normalize_...` in `utils`. I don't remember what we concluded, but IMO the latter is good so the check is really performed where needed, regardless on how we wrap or call the `normalize_...`. That way, we shouldn't need to change `laser.py`.

RemiLehe:
I updated the PR according to the above comment.

MaxThevenet (approved):
Looks good, thanks!

PR #241, opened 2024-04-19
Refactor speckled laser

RTSandberg:
This PR changes how the speckled lasers are structured and clarifies the documentation.

In particular, common functionality is provided in a base class and different forms of temporal smoothing can be expressed in child classes that only modify a few key functions

Questions/potential to-dos
- [x] Add temporal profile!
- [ ] Is there a way to protect the ``evaluate`` function of the base speckled class so children can't modify it?
- [ ] Is the parent documentation properly included in each child class documentation?
- [ ] Is ``*args`` the best way to send arguments to the parent ``__init__`` method?


RTSandberg:
I broke something in this PR and it was easier to rewind and start a new PR

PR #247, opened 2024-05-15
Allow import from openPMD file when rt-field has on axis grid node

soerenjalas:
Currently the code crashes with a rather cryptic interpolation error when importing an openPMD file that has data at r=0, because the function duplicates the first grid point and scipy doesn't like interpolation of non-strictly-monotonic data.

This PR should fix this behavior.


soerenjalas:
Just saw that @hightower8083 already proposed a fix in #245

PR #251, opened 2024-06-24
Add possibility to choose the number of points in theta

RemiLehe:
This PR adds a new, optional parameter `n_theta_evals` (see docstring in this PR.)

For highly asymmetrical lasers, setting this to a high number can be important in order to obtain a good approximation of the laser pulse we projecting it onto few azimuthal modes (i.e. symmetrizing it).


MaxThevenet (approved):

PR #255, opened 2024-07-16
Have multiple calls to `write_to_file` dump to different files

RemiLehe:
As pointed out by @soerenjalas, when calling `write_to_file` multiple times, the output was always written to the same file (the one corresponding to iteration `0`, in the openPMD convention.)

This PR fixes this by keeping track of the iteration number, and writing to separate iterations every time `write_to_file` is called.

**To be discussed:** Should we group all output files into a common folder? (like we do with PIC codes: typically this folder is called `diags` in that case.)

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -109,7 +109,7 @@ class Laser:
         self.grid = Grid(dim, lo, hi, npoints, n_azimuthal_modes)
         self.dim = dim
         self.profile = profile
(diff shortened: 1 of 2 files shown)


soerenjalas:
Thanks for the PR. Regarding your discussion point. If I recall correctly the user can in principle achieve this by themselves with `file_prefix` if that takes the form of a path, i.e. "diags/laser_". Although, this is admittedly not the most intuitive api.

soerenjalas (approved):

PR #256, opened 2024-07-16
Only perform spectral-to-temporal and temporal-to-spectral transforms when needed

RemiLehe:
Need to merge #199 first

This PR aims to improve efficiency by only perform spectral-to-temporal and temporal-to-spectral transforms when needed. This PR is meant to be an alternative to #227 

To this end, the `Grid` object keeps track of whether the validity of the spectral or temporal data, using getters and setters. The main changes to this PR are concentrated in the field `grid.py`. Here is a quick summary:
- Replaced the attribute `field` of the grid object by `temporal_field` and `spectral_field`
- Added `getter` and `setter` for both of these objects. This keeps track of the validity of the data. For instance, if a user uses the setter for `temporal_field`, the validity flag for `spectral_field` will be set to `False`, meaning that an FFT will be needed (from `temporal_field` to `spectral_field`) the next the user tries to access `spectral_field`.
- Also added a `shape` attribute that makes a few operations easier.

The rest of the PR adapts the `lasy` code to make use of the new interface.


lasy/laser.py
```diff
@@ -268,17 +309,6 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
         self.grid.hi[time_axis_indx] += translate_time
         self.grid.axes[time_axis_indx] += translate_time
 
-        # Translate the phase of spectral image
```
RemiLehe, line 317: This will need to be handled more carefully.

lasy/laser.py
```diff
@@ -213,6 +209,11 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
         distance : scalar
             Distance by which the laser pulse should be propagated
 
+        initial_optical_element: an :class:`.OpticalElement` object (optional)
+            Represents a thin optical element, through which the laser
+            propagates, before propagating for `distance` in free space.
+            If this is `None`, no optical element is used.
+
```
```suggestion
```

lasy/utils/grid.py
```diff
@@ -53,11 +55,111 @@ def __init__(self, dim, lo, hi, npoints, n_azimuthal_modes=None):
...
+        Set the temporal field.
+
+        Parameters
+        ----------
+        field : ndarray of complexs
+            The temporal field.
+        """
+        assert field.shape == self.temporal_field.shape
+        assert field.dtype == "complex128"
+        self.temporal_field = field
```
RemiLehe:
```suggestion
        self.temporal_field[:,:,:] = field
```

lasy/utils/grid.py
```diff
@@ -53,11 +55,111 @@ def __init__(self, dim, lo, hi, npoints, n_azimuthal_modes=None):
...
+        Set the spectral field.
+
+        Parameters
+        ----------
+        field : ndarray of complexs
+            The spectral field.
+        """
+        assert field.shape == self.spectral_field.shape
+        assert field.dtype == "complex128"
+        self.spectral_field = field
```
RemiLehe:
```suggestion
        self.spectral_field[:,:,:] = field
```

lasy/laser.py
```diff
@@ -10,6 +10,8 @@
 )
 from lasy.utils.openpmd_output import write_to_openpmd_file
 
+time_axis_indx = -1
```
```suggestion
from grid import time_axis_indx
```

lasy/utils/grid.py
```diff
@@ -1,9 +1,11 @@
 import numpy as np
 
+time_axis_indx = -1
+
 
 class Grid:
     """
-    Store an array (typically the envelope) and corresponding metadata.
+    Store the envelope in temporal and spectral space and corresponding metadata.
```
MaxThevenet, line 8: The new comment is better considering how LASY works, but at this point https://github.com/LASY-org/lasy/blob/72398acbacdfb9002e6a7e40e9a5adbda7aa2628/lasy/utils/laser_utils.py#L552 the grid does not represent the envelope. I think it's still better to keep your new comment here.

MaxThevenet (approved):
Looks great!

PR #264, opened 2024-07-16
[WIP] add a laser_energy attribute to Profile, to fix bug in SummedProfile

MaxThevenet:


MaxThevenet:
Fixing this in the code causes more confusion, so we will update the documentation instead.

PR #269, opened 2024-07-18
New resampling

NadezhdaKHACHAT:
Copy of a #202 PR but the library is mostly up to date. Added an automated test for the resampling feature.

Adds the option to resample the field onto a new grid during the propagation step. This can be useful for example when focusing the laser from a large beam size to a small spot.

The resampling option only works in the 'rt' geometry. In order to resample the fields, the user can pass a new grid object (changing the radial size of the grid but not the resolution) to the propagate() function. The new grid replaces the original grid.


RemiLehe:
@NadezhdaKHACHAT Thanks a lot for this PR!
Would you be able to fix the `git` conflicts here?

tests/test_resampling.py
```diff
@@ -0,0 +1,57 @@
...
+def test_resampling():
+    # Define the initial grid for the laser
+    dim = "rt"
+    lo = (0e-3, -90e-15)
+    hi = (15e-3, +90e-15)
+    npoints = (250, 30)
+
+    # Define the new grid for the laser
+    new_r_max = 300.e-6
+    new_grid = Grid(dim, lo, (new_r_max, hi[1]), npoints, n_azimuthal_modes=laser.grid.n_azimuthal_modes)
```
RemiLehe: `laser.grid.n_azimuthal_modes` is not defined at this point, which causes this test to fail for now.

lasy/laser.py
```diff
@@ -221,25 +223,52 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
         show_progress : bool (optional)
             Whether to show a progress bar when performing the computation
         """
+
```
```suggestion
```

lasy/laser.py
```diff
@@ -216,8 +218,13 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
             will be attenuated (to assert proper Hankel transform).
             Only used for ``'rt'``.
 
+        grid : Grid object (optional)
+            Resample the field onto a new grid of different radial size and/or different number
+            of raidal grid points. Only works for ``'rt'``.
```
```suggestion
            of radial grid points. Only works for ``'rt'``.
```

lasy/laser.py
```diff
@@ -237,33 +244,76 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
...
-                    overwrite=True,
+                    overwrite=False,
                     show_progress=show_progress,
                 )
-                spectral_field[i_m, :, :] = np.transpose(transform_data).copy()
-            self.grid.set_spectral_field(spectral_field)
+                if grid is not None:
+                    # Store the updated and resampled field.
+                    spectral_field_n[i_m, :, :] = np.transpose(transform_data).copy()
+                    spectral_field = spectral_field_n
```
RemiLehe, line 311: Could this be simplified to:
```suggestion
                    spectral_field[i_m, :, :] = np.transpose(transform_data).copy()
```

lasy/laser.py
```diff
@@ -237,33 +244,76 @@ def propagate(self, distance, nr_boundary=None, backend="NP", show_progress=True
...
+                    # Delete Propagator if resampling was done
+                    del self.prop
```
RemiLehe: This needs to be outside of the loop 
```
for i_m in range(self.grid.azimuthal_modes.size):
```
correct?

tests/test_resampling.py
```diff
@@ -0,0 +1,77 @@
+# -*- coding: utf-8 -*-
+"""Checking the implementation of the resampling propagator.
+
+Initializing a Gaussian pulse in the near field, and
+propagating it through a parabolic mirror, and then to
+the focal position (radial axis is resampled to accomodate the new size of the beam) ; we then check that the waist as the
```
RemiLehe:
```suggestion
the focal position (radial axis is resampled to accomodate the new size of the beam) ; we then check that the waist has the
```

RemiLehe (changes requested):
Thanks for this PR.

I noticed a potential bug, I think, in the case where several azimuthal modes are present. Could you include a case with multiple azimuthal modes (e.g. Laguerre-Gauss pulse focused by a parabolic mirror) in the automated test?

rob-shalloo:
Hey, this is a really great PR! Thanks @NadezhdaKHACHAT! I've been testing it out recently and one question is whether we should include some information in the doc string relating to the resolution and range of both grids and how they should relate to the propagation distance. As an example, if you take your test case and just change the focal length to 0.5 m (not crazy for a 5 mm beam), the numerical calculation no longer matches the theory prediction. 
Another option is to include some kind of helper function to flag when the chosen parameters are unlikely provide physical results. I think Igor did this in the original propagator (see `def check_new_grid(self, dz):`)

RemiLehe:
It looks like the option `overwrite=False` is causing issues.

For instance, the test `test_gaussian_propagator.py` is currently failing on my local machine, whereas it should not be affected by this PR.

When changing:
```python
                transform_data = self.prop[i_m].step(
                    transform_data,
                    distance,
                    overwrite=False,
                    show_progress=show_progress,
                )
```
to 
```
                transform_data = self.prop[i_m].step(
                    transform_data,
                    distance,
                    overwrite=True,
                    show_progress=show_progress,
                )
```
This test passes again.

RemiLehe:
It seems that the above-mentioned issue is fixed by this `axiprop` commit:
https://github.com/hightower8083/axiprop/commit/25d58eaf37941bba1274ad2011f1760fc8534fb1
which is not included in the latest PyPI release.

RemiLehe (approved):

PR #270, opened 2024-07-29
centralized setup of spectral axis

hightower8083:
Currently spectral axis `omega` is independently created in the separate `apply_optics` and `propagate` methods, which doubles the code. this PR centralizes this procedure 
 - the envelope spectral axis is created and registered along with the `grid` construction as soon as we have `dt` and `Nt`
 -  the actual spectral axis  is created and registered at the `laser` object

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -183,7 +183,9 @@ class Laser:
             # The line below assumes that amplitude_multiplier
             # is cylindrically symmetric, hence we pass
             # `r` as `x` and 0 as `y`
-            multiplier = optical_element.amplitude_multiplier(r, 0, omega, self.profile.omega0)
+            multiplier = optical_element.amplitude_multiplier(
+                r, 0, omega, self.profile.omega0
+            )
             # The azimuthal modes are the components of the Fourier transform
             # along theta (FT_theta). Because the multiplier is assumed to be
             # cylindrically symmetric (i.e. theta-independent):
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for this PR.
I would have a slight preference for storing `omega_1d` inside the `grid` object. This would require passing `omega0` to the `Grid` object when we create it.

RemiLehe:
Thanks for this PR.

I think it would make sense to avoid duplicating some of the properties (e.g. `omega0` being saved both in the `Laser` object and `Profile` object,  `omega_1d` and `omega_env` being almost the same, etc.)

I made some changes accordingly in this commit: https://github.com/LASY-org/lasy/pull/270/commits/4bb2914061646f744f176251aa6ffdaf0ea45231
Feel free to revert it if you prefer the previous version.

RemiLehe (approved):

hightower8083:
Hi @RemiLehe 

My logic was to create the variables as soon as the input is provided, i.e. as `grid` is created it can already have it's spectral axis regardless to the enveloping which comes in the laser. I can imagine that the `Grid` class may in future be reused with other classes than the `Laser`, for example, if we ever decide to make a non-enveloped class. In any case, now I see that it'd be more clear to use something like `omega_grid` instead of `omega_env`.

However, I'm not sure what should be the principle that we follow in `lasy` =).  If we prefer to avoid adding new variables to the class, I don't mind and we can always do extensions later if needed.

PR #275, opened 2024-08-05
Update documentation to discuss sum of profiles

MaxThevenet:
Supersedes #264.

While `Profile` and `TransverseProfile` classes have a `__add__` operator, @RemiLehe in particular noticed that summing different profiles could be risky. This PR proposes to update the documentation to be transparent with the user.


rob-shalloo (approved):
Thanks very much for this PR! I think its a great idea to add in some cautions regarding this topic. Happy to merge as is.

RemiLehe (approved):

PR #276, opened 2024-08-05
Add the axicon as a new optical element

cdmccombs:
added axicon optical element and axicon test

PR #278, opened 2024-08-08
added scaling test for axicon

cdmccombs:
Pass tests, however i think further discussion is necessary for the error test


RemiLehe:
These changes will be implemented in #276

PR #282, opened 2024-08-15
Remove superfluous `set_spectral_field`

RemiLehe:
This line is not needed, since we call this function further down.


MaxThevenet (approved):
Thanks!

PR #289, opened 2024-08-22
Remove member function of the `Grid` object

RemiLehe:
The member functions of `Grid` are meant to be used internally in `lasy` and should probably not show up in the user-facing documentation.


MaxThevenet (approved):
Agreed, thanks!

PR #296, opened 2024-09-18
Adding a tutorial to LASY

Paaaaarth:
Adding a tutorial to make a laser object using FROG data and transverse shot of the laser.


MaxThevenet (approved):
Thanks for this nice tutorial!

Issue #210, opened 2023-11-09
Error in reading tiff file 


alexanatoly:
Hi, I am trying to read an image file which is in a **.tiff** file format to test the example file (example_modal_decomposition_data.py). However, I am always getting an error 

modeCoeffs, waist = hermite_gauss_decomposition(

  File "/Users/labpc/opt/anaconda3/lib/python3.9/site-packages/lasy/utils/mode_decomposition.py", line 83, in hermite_gauss_decomposition
    w0 = estimate_best_HG_waist(x, y, field)

  File "/Users/labpc/opt/anaconda3/lib/python3.9/site-packages/lasy/utils/mode_decomposition.py", line 123, in estimate_best_HG_waist
    assert dx == dy
AssertionError

Any suggestions how to fix this error .....

RemiLehe:
Thanks for reporting this issue.

Would you be able to share your full Python script, so that we can reproduce the error?

Thanks,

alexanatoly:
Hi @RemiLehe , find attached the python script. For uploading, I included the extension '.txt' because '.py' files cannot be uploaded. Please remove '.txt' extension when using the file. 

As you can see, I just replace the image file name in the 'example_modal_decomposition_data.py' example file. The tiff image file is 16 MB in size. Files with size < 10 MB can be uploaded. If you need that also, I can send you by email.

[example_modal_decomposition_data_TIFF.py.txt](https://github.com/LASY-org/lasy/files/13312932/example_modal_decomposition_data_TIFF.py.txt)

alexanatoly:
Hi @RemiLehe, I used tiff file image because it is the most common file format for saving images in experiments. If the question is not in the scope of lasy, we can close this issue, however, if the error is within the scope of lasy and should be tested, I will keep it open.

RemiLehe:
@alexanatoly Sorry for the late reply! I think that your question is in the scope of `lasy`. 
In order for me to debug this further, could you please send me the image? You can find my email address on the LBNL phonebook: http://phonebook.lbl.gov/

alexanatoly:
@RemiLehe thank you. The tiff image file was sent to your email address.

alexanatoly:
Hi @RemiLehe, did you get time to see the tiff file ?
thanks.

MaxThevenet:
Should be fixed by #297.

alexanatoly:
Hi @MaxThevenet, 

Thanks for fixing this issue. I will give it a try. 

Alex

PR #297, opened 2024-09-19
[mini] Fix grid size calculation in mode decomposition

MaxThevenet:
The grid is calculated to have the exact resolution provided by the user. As before, the box width is approximately (but not exactly) `10*w0`, where `w0` is the spot size. The number of points is enforced to be even. This should address issue https://github.com/LASY-org/lasy/issues/210.

### DIFF
--- a/lasy/utils/mode_decomposition.py
+++ b/lasy/utils/mode_decomposition.py
@@ -67,15 +67,17 @@ def hermite_gauss_decomposition(laserProfile, n_x_max=12, n_y_max=12, res=1e-6):
         hi[0] = laserProfile.w0 * 5 + laserProfile.x_offset
         hi[1] = laserProfile.w0 * 5 + laserProfile.x_offset
 
-    N_pts_x = int((hi[0] - lo[0]) / res)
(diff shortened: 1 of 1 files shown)


lasy/utils/mode_decomposition.py
```diff
@@ -67,15 +67,23 @@ def hermite_gauss_decomposition(laserProfile, n_x_max=12, n_y_max=12, res=1e-6):
         hi[0] = laserProfile.w0 * 5 + laserProfile.x_offset
         hi[1] = laserProfile.w0 * 5 + laserProfile.x_offset
 
-    N_pts_x = int((hi[0] - lo[0]) / res)
-    N_pts_y = int((hi[1] - lo[1]) / res)
+    Nx = int((hi[0] - lo[0]) // (2 * res) * 2) + 2
```
MaxThevenet, line 70:
```suggestion
    # Number of grid points is even
    Nx = int((hi[0] - lo[0]) // (2 * res) * 2) + 2
```

RemiLehe (approved):

PR #300, opened 2024-10-04
Fix assert error when creating a grid from a field from file.

delaossa:
This addresses a problem when creating a lasy profile from a FBPIC output (OpenPMD) file.
It can be reproduced with the following piece of code:

```py

lasy_profile = FromOpenPMDProfile(
            path="lab_diags/hdf5",
            iteration=10,
            pol=(1, 0),  # dummy value, currently not needed
            field='E',
            coord='x',
            prefix=None,
            theta=0.0,
        )
```

The error comes when setting the complex, 3D array `grid.temporal_field` object in `grid.set_temporal_field` from the real, 2D array `field` read from file.
A solution to appropriately handling the array data type has been taken care of through #302.

This PR now simply fixes the dimensionality issue by adding an extra (azimuthal) dimension to the field array when passing it to `grid.set_temporal_field`.


MaxThevenet:
This is superseded by #311 who also ensures consistency, always making the arrays 3D.

Issue #301, opened 2024-10-04
Reading FBPIC output broken after release 0.4


MaxThevenet:
There is currently an issue that breaks our method to read FBPIC output in LASY. This breaks our FBPIC-LASY-WakeT pipeline, and people using it have to revert to LASY 0.4. I suspect the issue comes from #256. Ffrom an FBPIC output fails for 2 reasons, caught by asserts:
1. is `Grid.field` real or complex?
2. is it of shape [nr,nt] or [ntheta,nr,nt]?

@delaossa documents it and proposes a fix in #300. Additionally, for the first point, maybe we should store `is_envelope` as a member of `Grid` so we keep track of this easily? That would clean things up (we won't need to pass it as an additional argument in several `utils` functions) and make the code safer on this handling. I could give it a try.

PR #302, opened 2024-10-05
Make is_envelope a member of Grid

MaxThevenet:
Currently, `Grid` can be constructed with an array representing the full electric field (e.g. to use it to convert from field to envelope) or full envelope (as used internally in LASY). This PR proposes to clarify this behavior by making `is_envelope` a member variable of class `Grid`, assuming that the stored array is complex iff `is_envelope is True`. I see two advantages:
 - the use of `utils` like `field_to_envelope` would be safer and cleaner;
 - some other functions (e.g. propagator) could be applied to the full field, which would benefit from this handling.

This addresses issue #301 reported by @delaossa in #300.

### DIFF
--- a/lasy/utils/grid.py
+++ b/lasy/utils/grid.py
@@ -105,6 +105,7 @@ class Grid:
             The spectral field.
         """
         assert field.shape == self.spectral_field.shape
+        assert field.dtype == "complex128"
         self.spectral_field[:, :, :] = field
         self.spectral_field_valid = True
         self.temporal_field_valid = False  # Invalidates the temporal field
--- a/lasy/utils/grid.py
+++ b/lasy/utils/grid.py
@@ -67,7 +67,6 @@ class Grid:
             ncomp = 2 * self.n_azimuthal_modes - 1
             self.shape = (ncomp, self.npoints[0], self.npoints[1])
 
-        # This overwrites the dtype of temporal_field above
         self.set_is_envelope(is_envelope)
         self.temporal_field = np.zeros(self.shape, dtype=self.dtype)
         self.temporal_field_valid = False
--- a/lasy/utils/grid.py
+++ b/lasy/utils/grid.py
@@ -74,9 +74,7 @@ class Grid:
         self.spectral_field_valid = False
 
     def get_dtype(self):
-        """
-        Get the data type of self.temporal_field.
-        """
+        """Get the data type of self.temporal_field."""
         if self.is_envelope:
             return "complex128"
         else:
--- a/lasy/utils/grid.py
+++ b/lasy/utils/grid.py
@@ -81,6 +81,7 @@ class Grid:
             self.dtype = "complex128"
         else:
             self.dtype = "float64"
+        self.temporal_field = self.temporal_field.astype(dtype=self.dtype)
         self.is_envelope = is_envelope
 
     def set_temporal_field(self, field):
--- a/lasy/utils/grid.py
+++ b/lasy/utils/grid.py
@@ -69,8 +69,8 @@ class Grid:
             self.shape = (ncomp, self.npoints[0], self.npoints[1])
 
         self.temporal_field = np.zeros(self.shape, dtype=self.get_dtype())
-        self.spectral_field = np.zeros(self.shape, dtype=self.get_dtype())
         self.temporal_field_valid = False
+        self.spectral_field = np.zeros(self.shape, dtype="complex128")
         self.spectral_field_valid = False
 
     def get_dtype(self):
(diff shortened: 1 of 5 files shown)


lasy/utils/laser_utils.py
```diff
@@ -589,11 +583,11 @@ def field_to_envelope(grid, dim, phase_unwrap_nd=False):
     omg_h, omg0_h = get_frequency(
         grid,
         dim=dim,
-        is_envelope=False,
         is_hilbert=True,
         phase_unwrap_nd=phase_unwrap_nd,
     )
     field *= np.exp(1j * omg0_h * grid.axes[-1])
+    grid.is_envelope = True
```
delaossa: I think that we also need to change the data type of `grid.temporal_field` to  `complex128`.
If not, we get a `complex` to `float` conversion warning in `grid.set_temporal_field`.
```
ComplexWarning: Casting complex values to real discards the imaginary part
  self.temporal_field[:, :, :] = field
 ```

**Review by delaossa — COMMENTED** [review]
I think that there is a missing bit when changing `grid.is_envelope` to True. The data type of `grid.temporal_field` needs to be changed accordingly to `complex` to avoid type conversion warnings when setting the new field.

Suggestion: we can add a function to the grid class to perform this task.
I have implemented this in the other PR (which I already merged with this one)
https://github.com/LASY-org/lasy/pull/300/commits/33c7af075f74aad7fbf4e5989da173fbf9f8c1d1

File: lasy/utils/grid.py [hunk 2]
```diff
@@ -88,7 +88,8 @@ def set_is_envelope(self, is_envelope):
             self.dtype = "complex128"
         else:
             self.dtype = "float64"
-        self.temporal_field = self.temporal_field.astype(dtype=self.dtype)
+        if hasattr(self, "temporal_field"):
+            self.temporal_field = self.temporal_field.astype(dtype=self.dtype)
```
delaossa, line 91: This avoids unnecessary casting from complex to float at initialization with `is_envelope = False`

lasy/utils/grid.py
```diff
@@ -67,9 +67,9 @@ def __init__(self, dim, lo, hi, npoints, n_azimuthal_modes=None, is_envelope=Tru
             ncomp = 2 * self.n_azimuthal_modes - 1
             self.shape = (ncomp, self.npoints[0], self.npoints[1])
 
-        self.temporal_field = np.zeros(self.shape, dtype="complex128")
         # This overwrites the dtype of temporal_field above
         self.set_is_envelope(is_envelope)
+        self.temporal_field = np.zeros(self.shape, dtype=self.dtype)
```
delaossa: This reordering conveniently initialize the array after the type has been selected.
We can now delete the comment in line 70.

delaossa (approved):

lasy/utils/grid.py
MaxThevenet: Indeed, thanks for the comment!

lasy/utils/grid.py
```diff
@@ -62,11 +67,31 @@ def __init__(self, dim, lo, hi, npoints, n_azimuthal_modes=None):
             ncomp = 2 * self.n_azimuthal_modes - 1
             self.shape = (ncomp, self.npoints[0], self.npoints[1])
 
-        self.temporal_field = np.zeros(self.shape, dtype="complex128")
+        # This overwrites the dtype of temporal_field above
```
delaossa:
```suggestion
```
This comment is not needed anymore.

lasy/utils/laser_utils.py
```diff
@@ -279,14 +281,8 @@ def get_spectrum(
...
 
-    is_envelope : bool (optional)
-        Whether the field provided uses the envelope representation, as used
-        internally in lasy. If False, field is assumed to represent the
-        the full electric field (with fast oscillations).
-
     omega0 : scalar (optional)
-        Angular frequency at which the envelope is defined. Required if
-        `is_envelope=True`.
+        Angular frequency at which the envelope is defined.
```
MaxThevenet, line 285:
```suggestion
        Angular frequency at which the envelope is defined.
        Only used if grid.is_envelope is True.
```

lasy/utils/laser_utils.py
```diff
@@ -379,27 +375,21 @@ def get_frequency(
...
-        internally in lasy. If False, field is assumed to represent the
-        the electric field.
-
     is_hilbert : boolean (optional)
         If True, the field argument is assumed to be a Hilbert transform, and
         is used through the computation. Otherwise, the Hilbert transform is
         calculated in the function.
 
     omega0 : scalar
         Angular frequency at which the envelope is defined.
```
MaxThevenet, line 393:
```suggestion
        Angular frequency at which the envelope is defined.
        Only used if grid.is_envelope is True.
```

rob-shalloo (approved):
Hey thanks for this PR, I think this is certainly a cleaner way of doing things and will ultimately make the code safer in its handling of both envelope and full field data

Issue #185, opened 2023-10-09
laser wavelength must be monotonically increasing in `LongitudinalProfileFromData`


rob-shalloo:
Have just come across a bug when using the `LongitudinalProfileFromData` profile in the `spectral` domain that the wavelength must be monotonically increasing. If the spectrum extends for example from 1000 nm to 500 nm rather than 500 nm to 1000 nm, then the function fails to produce something sensible.

The fix should be relatively straightforward by checking this condition and either adding an assert or flipping the array

PR #304, opened 2024-10-08
Added tests for LongitudinalProfileFromData and fix for issue #185 

em-archer:
Editing LongitudinalProfileFromData to allow for monotonically decreasing spectral data (fix for #185) and adding relevant tests to test_laser_profiles. Currently no tests exist for LongitudinalProfileFromData, so adding tests for monotonically increasing/decreasing spectral data. Test for temporal data could be added in future.

### DIFF
--- a/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
+++ b/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
@@ -63,14 +63,16 @@ class LongitudinalProfileFromData(LongitudinalProfile):
 
     def __init__(self, data, lo, hi):
         if data["datatype"] == "spectral":
-            if not "axis_is_wavelength" in data: # Set to wavelength data by default
+            if "axis_is_wavelength" not in data:  # Set to wavelength data by default
                 data["axis_is_wavelength"] = True
             if data["axis_is_wavelength"]:
                 wavelength = data["axis"]  # Accept as wavelength
                 spectral_intensity = data["intensity"]
             else:
                 wavelength = 2.0 * np.pi * c / data["axis"]  # Convert to wavelength
-                spectral_intensity = data["intensity"] * 2.0 * np.pi * c / wavelength ** 2 # Convert spectral data
+                spectral_intensity = (
+                    data["intensity"] * 2.0 * np.pi * c / wavelength**2
+                )  # Convert spectral data
             assert np.all(np.diff(wavelength) > 0) or np.all(
                 np.diff(wavelength) < 0
             ), 'data["axis"] must be in monotonically increasing or decreasing order.'
--- a/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
+++ b/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
@@ -65,18 +65,15 @@ class LongitudinalProfileFromData(LongitudinalProfile):
             assert np.all(np.diff(wavelength) > 0) or np.all(
                 np.diff(wavelength) < 0
             ), 'data["axis"] must be in monotonically increasing or decreasing order.'
-            if np.all(np.diff(wavelength) < 0):
-                wavelength = wavelength[::-1]
-                spectral_intensity = data["intensity"][::-1]
-            else:
-                spectral_intensity = data["intensity"]
+            spectral_intensity = data["intensity"]
             if data.get("phase") is None:
                 spectral_phase = np.zeros_like(wavelength)
             else:
-                if np.all(np.diff(wavelength) < 0):
-                    spectral_phase = data["phase"][::-1]
-                else:
-                    spectral_phase = data["phase"]
+                spectral_phase = data["phase"]
+            if np.all(np.diff(wavelength) < 0: # Flip arrays
+                wavelength = wavelength[::-1]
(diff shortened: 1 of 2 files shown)


em-archer:
Also added small update to also accept angular spectrum data (units of s^-1) instead of wavelength data

tests/test_laser_profiles.py
```diff
@@ -234,6 +238,89 @@ def test_longitudinal_profiles():
     print("cep_phase = ", cep_phase_cos)
     assert np.abs(cep_phase_cos - cep_phase) / cep_phase < 0.02
 
+    # LongitudinalProfileFromData
+    print("LongitudinalProfileFromData")
+    data = {}  # Generate spectral data assuming analytic ft of GaussianLongitudinalProfile
```
MaxThevenet: Is this correct? Just to make it more explicit.
```suggestion
    data = {}  # Generate spectral data assuming analytic Fourier transform of GaussianLongitudinalProfile
```

MaxThevenet (commented):
Thanks for this PR! Really great to have more tests, and the option to initialise from file while passing the angular frequency instead of wavelength will be useful, in particular for @Paaaaarth in PR #296. Besides a minor suggestion in the text, I have two comments:
1. Determining whether it is wavelength or frequency depending on the value is IMO not very straightforward. Would it be possible to make the user provide specific arguments? The user could provide `datatype = spectral` as before, but there could be one additional argument `axis_is_wavelength` with default `True`, and if set to `False` the axis is supposed to be frequency. I think this parameter would appear clearly in the doc and should make it clear to the users.
2. When converting from axis = frequency to wavelength, shouldn't we also modify the spectrum by multiplying with $2\pi c / \lambda^2$ or so? Not sure this is done in the current reader, could you check this?

lasy/profiles/longitudinal/longitudinal_profile_from_data.py
```diff
@@ -60,16 +63,27 @@ class LongitudinalProfileFromData(LongitudinalProfile):
...
     def __init__(self, data, lo, hi):
         if data["datatype"] == "spectral":
-            # First find central frequency
-            wavelength = data["axis"]
-            assert np.all(
-                np.diff(wavelength) > 0
-            ), 'data["axis"] must be in monotonously increasing order.'
-            spectral_intensity = data["intensity"]
+            if "axis_is_wavelength" not in data:  # Set to wavelength data by default
+                data["axis_is_wavelength"] = True
```
MaxThevenet, line 69: I think this will modify the actual `data` object passed by the user: if the user defines an actual `data` variable without `"axis_is_wavelength"` and passes it to this function, it will have this attribute defined after this call.
```py
data = ... # "axis_is_wavelength" not in data
p = LongitudinalProfileFromData(data, ...)
# now "axis_is_wavelength" is in data
```
In most cases, this will have absolutely no importance. But I think it is still safer to avoid modifying it. The solution is simple, define a new variable, and use this one instead, e.g.
```py
if "axis_is_wavelength" not in data:
    axis_is_wavelength = True
else:
    axis_is_wavelength = data["axis_is_wavelength"]
```

lasy/profiles/longitudinal/longitudinal_profile_from_data.py
```diff
@@ -25,12 +25,15 @@ class LongitudinalProfileFromData(LongitudinalProfile):
 
         datatype : string
             The domain in which the data has been passed. Options
-            are 'spectral' and 'temporal'
+            are 'spectral' and 'temporal'. Spectral data may either be,
+            a) Wavelength dependent (lambda): data["axis_is_wavelength"] = True or
+            b) Angular freuqency dependent (omega): data["axis_is_wavelength"] = False
```
MaxThevenet: Considering suggestion below.
```suggestion
            are 'spectral' and 'temporal'.
```

lasy/profiles/longitudinal/longitudinal_profile_from_data.py
```diff
@@ -25,12 +25,15 @@ class LongitudinalProfileFromData(LongitudinalProfile):
...
```
MaxThevenet, line 33: This option probably warrants a separate entry, here's a suggestion for this.
```suggestion

        axis_is_wavelength : boolean (optional, default True)
            If True, the axis represents wavelength in [m] (SI).
            If False, it represents frequency in [1/s] (SI).

```

lasy/profiles/longitudinal/longitudinal_profile_from_data.py
```diff
@@ -25,12 +25,15 @@ class LongitudinalProfileFromData(LongitudinalProfile):
...
 
         axis : ndarrays of floats
             The horizontal axis of the pulse duration measurement.
-            The array must be monotonously increasing.
-            When datatype is 'spectral' axis is wavelength in meters.
+            The array must be monotonically increasing or decreasing.
+            When datatype is 'spectral' axis is wavelength in meters OR
+            angular frequency in 1/seconds.
```
MaxThevenet, line 38:
```suggestion
            angular frequency in 1/seconds.
```

MaxThevenet (commented):
Thanks! See a few comments below, I think that's the last round.

MaxThevenet:
Superseded by #309.

PR #311, opened 2024-10-21
More check on number of dimensions for rt

MaxThevenet:
LASY should always initialise a `Grid` with a 3D array, either `(x, y, t)` or `(mode, r, t)`. This PR adds an `assert` accordingly and fixes one `create_grid` call so it complies with this.

This should address #301, similar to #300 but doing the fix in a different place. @delaossa could you check that this also fixes the issue you reported on #300?

### DIFF
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -95,8 +95,14 @@ class FromOpenPMDProfile(FromArrayProfile):
             dim = "rt"
             axes_order = ["r", "t"]
 
+        # Set r and t axes to the right dimension for LASY.
+        # Note that F is assumed 2D here, no azimuthal data.
         F, axes = reorder_array(F, m, dim)
 
+        if dim == 'rt' and F.ndim == 2:
+            F = F[np.newaxis, :]
+        elif F.ndim != 3:
+             raise ValueError(f"F has the wrong number of dimensions: {F.ndim} (should be 3)")
         # If array does not contain the envelope but the electric field,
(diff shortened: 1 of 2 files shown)


rob-shalloo:
Hi, thanks for this PR. I think that clarifying this will be very helpful.

One question I would have is whether this change to always initialise a 3D array should be reflected in the `dim` variable throughout the code. As an example of what I'm thinking of, I would consider relabelling `dim='rt'` to `dim='mrt'` throughout the codebase. For me this would help improve transparency, but of course, also has its own implications for example in the definition of `lo` and `hi`.

MaxThevenet:
Good point. It is true that in a number of places, for instance in this function
https://github.com/LASY-org/lasy/blob/89facdbfc464e58c6b24a018957b90f56322cab1/lasy/utils/laser_utils.py#L660
we have some variables like lo and hi, in this case axes, that do [r,t] while the array is [m,r,t]. This is a bit more straightforward in 3D, where these are [x,y,t] consistent with the array. Should we move this to a separate issue? Note that this would be a non-backward-compatible change, but at least would not affect the standard I think.

delaossa:
> This should address #301, similar to #300 but doing the fix in a different place. @delaossa could you check that this also fixes the issue you reported on #300?

Yes, this fixes the issue reported on #300.
Thanks @MaxThevenet !

delaossa (approved):
Looks good. Thanks!

hightower8083 (approved):
looks good to me

RemiLehe (approved):

PR #315, opened 2024-10-23
Utility to denoise laser profile and adding a test for the same

Paaaaarth:
Denoising can be considered as one of the common thing that's done while creating a laser object and yet there is no trivial way to do so in LASY. With this pull request we aim to add this functionality i.e. mode decomposition based denoising. In its current state the code does two distinct things, mode decomposition and creating denoised profile.


MaxThevenet (changes requested):
Thanks for this PR! A few general points:
1. Good idea to put this in a separate file `denoise.py` as you propose, if we consider having other denoising methods implemented later. Could you adapt the naming etc.? For instance, the function could be called: `denoise_LG_modes` or so.
2. I would encourage you to check [`mode_decomposition.py`](https://github.com/LASY-org/lasy/blob/development/lasy/utils/mode_decomposition.py), it can take either a full Laser profile or a Transverse profile, and explain what it does in each case. IMO the clearest would be for your function to take a generic `profile` that can be a `TransverseProfile` or a `Laser` but always return a `TransverseProfile`. In the future, we could always return the same type as input, I think that would be the best, but would take more work to handle the longitudinal data.
3. Could you avoid ambiguous input arguments like `parameters` bundling different variables, unless necessary? Just pass the arguments you need, and no more (for instance you may get rid of `polarization`).
4. Please document all input parameters, check other functions for example.
5. Please add a CI test.

Paaaaarth:
With D4, We have added the additional capability of inputing an image file as transverse profile. Skimage package is used to extract the intensity data.

rob-shalloo:
Hey @Paaaaarth thanks for the PR! This is great to have the functionality in one place. I would make a few suggestions. Feel free to take or leave as you see fit.

1) While this is a great technique to denoise laser profiles, I actually think this goes much beyond just denoising. We're really talking here about creating an analytic representation of the laser pulse (as a sum of HG modes) from an arbitrary input profile. I would then suggest that you change the name to something like `hg_reconstruction`. We could of course collect such methods as Maxence suggest like `lg_reconstruction` in one place within the documentation to showcase their use for denoising.

2) I would suggest removing the creation of a `transverse_profile_from_data` here and leave that to the user. Ie have an assert that you only accept existing transverse profiles or again as Maxence suggest full laser profiles could be interesting in the future. I think the `transverse_profile_from_data` complicates things as it would not be so generic. You in any case do not allow the user to select how the raw image is processed, for example, the threshold at which you set pixel values to 0 (here it appears to be anything less than 1%). This always takes some tuning depending on the camera model, bit depth, background subtraction, signal strength etc.

3) This is probably a larger 'fix' but I would also suggest we allow different values of the waist in `x` and `y` and actually that in addition to providing some functionality to "guess" the values, we allow the user to specify values. A sensible choice of waist can quite often speed up the decomposition / reconstruction process. This would mean digging into `hermite_gauss_decomposition` a bit.

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+    n_modes_y=2,
+):
+    """
+    Denoise the transverse profile by decomposing it into a set of Hermite-Gaussian modes.
+
+    The profiles are weighted according to mode coefficients and then added.
+
+    Parameters
+    ----------
+    transverse_profile : class instance
```
MaxThevenet:
```suggestion
    transverse_profile : TransverseProfile object
```

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+        An instance of a class or sub-class of TransverseProfile. Or it could be
+        a path to an image file from which intensity data will be read.
```
MaxThevenet: Let avoid ambiguity in the input parameter: it has one type, simple. In order to read an image, the user would create their own `TransverseProfileFromData` and pass it there ok?
```suggestion
        An instance of a class or sub-class of TransverseProfile.
```

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+        Defines the transverse envelope of the laser.
+
+    wavelength : float (in meter)
+        Central wavelength at which the Hermite-Gauss beams are to be defined.
+
+    resolution : float
+        The resolution of grid points in x and y that will be used
+        during the decomposition calculation.
```
MaxThevenet:
```suggestion
        The number of grid points in x and y used during the decomposition calculation.
```

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+
+    lo, hi : array of floats
+        The lower and upper bounds of the spatial grid on which the
+        decomposition will be performed.
```
MaxThevenet: Overall I personally would have a preference for present tense all over (there are some more instances below).
```suggestion
        decomposition is be performed.
```

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+    ----------
+    transverse_profile : class instance
+        An instance of a class or sub-class of TransverseProfile. Or it could be
+        a path to an image file from which intensity data will be read.
+        Defines the transverse envelope of the laser.
+
+    wavelength : float (in meter)
+        Central wavelength at which the Hermite-Gauss beams are to be defined.
+
+    resolution : float
```
MaxThevenet, line 31: Please specify here too when a parameter is optional. Also true for others below.

lasy/utils/denoise.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def denoise_transverse_hg(
+    transverse_profile,
+    wavelength,
+    resolution=0.2e-6,
+    lo=[-2e-4, -2e-4],
+    hi=[2e-4, 2e-4],
+    n_modes_x=2,
+    n_modes_y=2,
```
MaxThevenet: I think only `n_modes_x/y` should be optional, and given a meaningful value, e.g., 10. For the others, the user will get something likely wrong and confusing with default values.

MaxThevenet (changes requested):
Thanks for this PR! The core of it looks good, see some comments below.
As per Rob's points:
1. the naming: sounds good to me, I don't do denoising etc. on a daily basis.
2. agreed, I think I have a similar comment in a comment below.
3. Indeed the option of having 2 waists would be nice. Can be done in a follow-up PR.

tests/test_denoise.py
```diff
@@ -0,0 +1,40 @@
...
+        transverse_profile, wavelength, resolution, lo, hi
+    )  # Denoised profile
+
+    # Calculate the error
+    x = np.linspace(-5 * w0x, 5 * w0x, 500)
+    X, Y = np.meshgrid(x, x)
+    prof1 = np.abs(transverse_profile.evaluate(X, Y)) ** 2
+    prof2 = np.abs(transverse_profile_cleaned.evaluate(X, Y)) ** 2
+    error = (prof1 - prof2) / np.max(prof1)
+    assert 0 < np.max(error) < 0.2
```
MaxThevenet: This condition is IMO a bit weak. Could you make it a bit stronger? A different norm as used below (code not tested) would be preferable.
```suggestion
    prof1 = transverse_profile.evaluate(X, Y)
    prof2 = transverse_profile_cleaned.evaluate(X, Y)
    error = np.sum(np.abs(prof2-prof1)**2) / np.sum(np.abs(prof1)**2)
    assert error < 0.01 # or something
```

MaxThevenet (changes requested):
Looks great! See 1 comment below.

MaxThevenet:
Could you address the comment before requesting another review?

Paaaaarth:
> Could you address the comment before requesting another review?

Requested change has been implemented:)

MaxThevenet (approved):
Looks great, thanks!

PR #263, opened 2024-07-16
Polynomial Spectral Phase element

soerenjalas:
Merge #261 first! 

This PR adds an optical element that allows to add spectral phase parametrized through the Taylor series coefficients GDD, TOD, and FOD.

Description edited by @RemiLehe:

I added two automated test:

- When adding GDD to a Gaussian, the pulse broadens and acquires a chirp (which, in the envelope representation, corresponds to variations of the real and imaginary parts). This can be predicted analytically:
Screenshot 2024-07-25 at 2 50 44 AM

- When adding TOD>0 to a Gaussian, it creates post-pulses, for which the amplitude can be predicted analytically, using the stationary phase approximation:
Screenshot 2024-07-25 at 3 17 35 AM

PR #323, opened 2024-11-18
Remove `omega0` from explicit inclusion in `optical_element`

rob-shalloo:
I've made a **suggested** modification to the `optical_element` class to remove `omega0` from explicit inclusion in the method `amplitude_multiplier`. 

I find it somewhat confusing that we have both `omega` and `omega0` as input arguments. Seeing them both, I would automatically assume that the frequency axis `omega` is then actually centered around 0, and that `omega` is rather $\Delta \omega$ such that for each optical element we would need to use `omega + omega0` but this is not the case. Instead all but one of our optical elements do not use `omega0` at all.

I understand that this was specifically introduced in #261 for the case of polynomial spectral phase #263 . Here I believe we should instead add `omega0` as an argument required for initialisation of this optic. This optic adds spectral phase as a polynomial expansion of frequency and so defining the central frequency explicitly is, I think, a good idea.

In any case, as mentioned this is just a suggestion so totally fine to not make the change if people prefer it the way it is. Everything mechanically works great as is

@soerenjalas , could you take a look and see what you think, given you created the two PRs above

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -183,9 +183,7 @@ class Laser:
             # The line below assumes that amplitude_multiplier
             # is cylindrically symmetric, hence we pass
             # `r` as `x` and 0 as `y`
-            multiplier = optical_element.amplitude_multiplier(
-                r, 0, omega
-            )
+            multiplier = optical_element.amplitude_multiplier(r, 0, omega)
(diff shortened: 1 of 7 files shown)


MaxThevenet (approved):
Thanks! That looks good to me.
@soerenjalas any comment?

PR #325, opened 2024-11-19
Update axicon.py

David1one:
Fix documentation - missing factor of c


RemiLehe (approved):
Thanks a lot for pointing out and fixing this!

PR #326, opened 2024-11-21
Correction to the mode decomposition

Paaaaarth:
We have opened this pull request to modify the mode decomposition in order to make it more versatile. We have removed the dependence on field_interp which is a function of transverse profile from data class and not of the base transverse profile class.


lasy/utils/mode_decomposition.py
```diff
@@ -9,7 +9,15 @@
...
 
-def hermite_gauss_decomposition(laserProfile, wavelength, m_max=12, n_max=12, res=1e-6):
+def hermite_gauss_decomposition(
+    laserProfile,
+    wavelength,
+    m_max=12,
+    n_max=12,
+    res=1e-6,
+    lo=[-2e-4, -2e-4],
+    hi=[2e-4, 2e-4],
```
MaxThevenet: Could you make these 3 mandatory parameters?

MaxThevenet (changes requested):
Thanks for this PR! See comment below.

lasy/utils/mode_decomposition.py
Paaaaarth: Thanks you for the feedback. The changes suggested have been made.

Paaaaarth: Thank you for the feedback:)
The requested changes have been made. Additionally n_max and m_max also needed to be made mandatory in order to keep the code as consistent as possible.

lasy/utils/mode_decomposition.py
```diff
@@ -9,7 +9,15 @@
 from lasy.utils.exp_data_utils import find_d4sigma
 
 
-def hermite_gauss_decomposition(laserProfile, wavelength, m_max=12, n_max=12, res=1e-6):
+def hermite_gauss_decomposition(
+    laserProfile,
+    wavelength,
+    m_max,
+    n_max,
```
MaxThevenet: I think it's good to keep the default 12 for these two, actually, people may wonder what makes sense. That indeed requires one change in the order of parameters in the doc.

MaxThevenet (changes requested):

lasy/utils/mode_decomposition.py
Paaaaarth: Requested changes have been made:)

MaxThevenet (approved):
Thanks!

PR #307, opened 2024-10-15
Enable initialisation and estimation of optional spatio-temperal coupling parameters to gaussian profile

huixingjian:
### PR Description

This PR, synergized with PR #182, introduces the initialization and estimation of first-order spatio-temporal coupling parameters in the ```GaussianLongitudinalProfile``` class. Four additional parameters have been added to the constructor:

- Group-delay dispersion `phi2` in $s^2$ parameterized by: $\varphi^{(2)} = \frac{dt_0}{d\omega}$, 
- Spatial chirp `zeta` in $m \\cdot s$ parameterized by: $\zeta = \frac{dx_0}{d\omega}$(Here $x_0$ is the beam center position), 
- Angular dispersion `beta` in $s$  parameterized by: $\beta = \frac{d\theta_0}{d\omega}$( Here $\theta_0$ is the propagation angle of this component.)
- Finally `stc_theta` is the direction of transverse spatio-coupling in xoy plane

### Issue addressed

With the introduction of spatio-temporal coupling (STC), the ```LongitudinalProfile``` class can leverage additional information from the transverse profile. Along with the aforementioned parameters, the waist and focus length are also incorporated into the constructor. The transverse points x and y have been set to default values of 0 in the evaluate member function.

In the laser utilities, a new function, ```get_STC```, has been added to evaluate the STC parameters using the following formulas, with the laser envelope expressed as $a = a_0 e^{i\theta}$:
- The temporal chirp is calculated through
  $\\phi^{(2)} = \\frac{4\\varphi^{(2)}}{4(\\varphi^{(2)})^2+\\tau^4}$
  Here $\\tau$ is duration in **_s_**, and `phi2` refers to group-delay dispersion $\\phi^{(2)} $, and $\\Phi^{(2)}$ can be calculated by $\\frac{\\partial^2 \\theta }{\\partial t ^2}$.
- Similarly, the spatial chirp is tested through:
  $\\nu = \\frac{4\\zeta }{w_0^2\\tau^2+4\\zeta^2}$
  Here $L_0$ and $w_0$ are the laser duration and laser waist respectively. $\zeta = \frac{dr_0}{d\omega}$, where $r_0$  is the transverse beam centroid ($\\theta$ is the ```stc_theta```mentioned above). 
- Finally, the angular chirp term is tested through:
  $\\beta = \\frac{d_{\\theta_0}}{ d\\omega}$
  with $\\theta_0$ refering to the propagating angle of the frequency component

### Benefits

[...]

PR #335, opened 2024-12-19
update show function to include intensity

rob-shalloo:
Requires merging of #307 first

modified the existing show method of the laser class to scale the temporal and spatial axes to commonly used units (ns,ps,fs or um, mm, m). 

Additionally, added a boolean to flag whether the user wants intensity plotted instead of amplitude. If Intensity is selected for plotting then spatial and temporal lineouts are also added together with simple measurements of the pulse duration and spot size. These measurements rely on functions included in #307 

Example of use:
```python
# Define the physical laser parameters
wavelength = 800e-9         # Laser wavelength in meters
polarization = (1, 0)       # Linearly polarized in the x direction
energy = 1.5                # Energy of the laser pulse in joules
spot_size = 25e-6           # Waist of the laser pulse in meters
pulse_duration = 30e-15     # Pulse duration of the laser in seconds
t_peak = 0.0                # Location of the peak of the laser pulse in time


# Create a laser profile which gathers all the physical properties of the laser
laser_profile = GaussianProfile(
    wavelength, polarization, energy, spot_size, pulse_duration, t_peak
)

# Evaluate the laser field on a computational grid
dimensions = "rt"                           # Use cylindrical geometry
lo = (0, -5 * pulse_duration)             # Lower bounds of the simulation box
hi = (5 * spot_size, 5 * pulse_duration)  # Upper bounds of the simulation box
num_points = (300, 500)                     # Number of points in each dimension

laser = Laser(dimensions, lo, hi, num_points, laser_profile)

laser.show(show_intensity=True)
```
Screenshot 2024-12-19 at 11 05 20


MaxThevenet:
Thanks for this PR! Could the implementation be moved to `utils` somewhere to keep the `Laser` class simple and lean? I think the call `show(laser,show_intensity=True)` would work great.

MaxThevenet (approved):
Looks great, thanks!

PR #338, opened 2024-12-19
Fix normalize intensity

kale-j:
Fixes bug introduced in commit 350059b
normalize_peak_intensity was attempting to set the field to a single number. The field should be multiplied by this factor instead, as it was prior to 350059b


RemiLehe (approved):

RemiLehe:
Thanks a lot for finding and fixing this!

PR #339, opened 2024-12-20
Add option to normalize using average intensity

kale-j:
Adds option to normalize the laser using the average intensity of the envelope. This is useful, for example, with speckled laser pulses.

Also includes tests for the different normalization options.


MaxThevenet (approved):
Thanks for this PR!

PR #341, opened 2024-12-28
Modify mode decomposition to gain independent value of waist in x and y dimension

Paaaaarth:
Merge before #315

The following pull request aims to modify mode decomposition in a way that the waist is no longer a simple circle, but has independent values in x and y direction. The task is performed in to make utility consistent with the rest of the code, in light of recent changes to HermiteGaussianTransverseProfile.

### DIFF
--- a/docs/source/tutorials/denoised_laser.ipynb
+++ b/docs/source/tutorials/denoised_laser.ipynb
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "dcf32943-cf31-42c3-995c-2ed75c9761d2",
    "metadata": {},
    "outputs": [],
@@ -48,7 +48,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "id": "e4a2fd86-a83f-490e-9e10-21d0e3ca7c4d",
    "metadata": {},
    "outputs": [],
@@ -84,7 +84,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "id": "7ee0575d",
    "metadata": {},
    "outputs": [],
@@ -109,7 +109,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
    "id": "e70f703e-15a3-4751-a857-71b17651671a",
    "metadata": {},
    "outputs": [],
@@ -226,7 +226,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 7,
    "id": "ac17f984",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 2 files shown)


docs/source/tutorials/denoised_laser.ipynb
```diff
@@ -516,7 +516,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.2"
+   "version": "3.11.4"
```
MaxThevenet:
```suggestion
   "version": "3.12.2"
```

lasy/utils/mode_decomposition.py
```diff
@@ -126,8 +127,8 @@ def estimate_best_HG_waist(x, y, field, wavelength):
 
     Returns
     -------
-    w0 : scalar
-        The calculated waist.
+    w0 : array of floats
+        The calculated waist in x and y axis.
```
MaxThevenet: I don't think one array of 2 elements is really appropriate here, for a few reasons:
- this is not consistent with x and y, passed as 2 different input arguments, and this is true for a few other variables;
- line 154 is very confusing, as x and y are treated asymmetrically in `wTest, waistTestY[I]` for no good reason;
- What you return is actually not array, rather a list.

Could you return e.g. `w0x` and `w0y` instead?

lasy/utils/mode_decomposition.py
```diff
@@ -136,20 +137,28 @@ def estimate_best_HG_waist(x, y, field, wavelength):
...
+    waistTestY = np.linspace(w0EstY / 2, w0EstY * 1.5, 30)
+    coeffTest = np.zeros_like(waistTestX)
 
-    for i, wTest in enumerate(waistTest):
+    for i, wTest in enumerate(waistTestX):
         # create a gaussian
-        HGMode = HermiteGaussianTransverseProfile(wTest, wTest, 0, 0, wavelength)
+        HGMode = HermiteGaussianTransverseProfile(
+            wTest, waistTestY[i], 0, 0, wavelength
+        )
```
MaxThevenet: Here, x and y are not treated symmetrically, which is not justified.

MaxThevenet (changes requested):
Thanks for this PR! See mostly one comment below.

lasy/utils/mode_decomposition.py
Paaaaarth: All the requested changes have been implemented:)

lasy/utils/mode_decomposition.py
Paaaaarth: All the requested changes have been implemented:)

lasy/utils/mode_decomposition.py
```diff
@@ -148,17 +148,16 @@ def estimate_best_HG_waist(x, y, field, wavelength):
     waistTestY = np.linspace(w0EstY / 2, w0EstY * 1.5, 30)
     coeffTest = np.zeros_like(waistTestX)
 
-    for i, wTest in enumerate(waistTestX):
+    for i in range(0, 30):
```
MaxThevenet:
```suggestion
    for i in range(30):
```

MaxThevenet (approved):

PR #349, opened 2025-02-06
Implementation of Flattened-Gaussian transverse profile

delaossa:
Copied from FBPIC :) -> https://github.com/fbpic/fbpic/blob/dev/fbpic/lpa_utils/laser/transverse_laser_profiles.py

I have tested that the analytical profile far from focus (initialized with `z_foc`) matches the profile obtained with `propagate` at the same distance from focus.

Using `z_foc = 4 * zR`
laser_profile

Using `z_foc = 0` and laser.propagate(- 4 * zR)
laser_profile_prop


MaxThevenet (approved):
Looks great, thanks!

PR #356, opened 2025-02-14
Allows reading laser envelope from Wake-T

delaossa:
Reading Wake-T files with `FromOpenPMDProfile` was not working well.
This PR includes a patch to make it work. 
It may serve as a basis to discuss a better and more general implementation of `FromOpenPMDProfile` and `FromArrayProfile` classes.

- A first issue was in the way that this method access the `angularFrequency` attribute in the hdf5 file, which is very specific for `LASY` envelope files only. Now, a new variable `is_waket` can be passed to `FromOpenPMDProfile` to read the Wake-T file. We should generalize the method. In the meantime this serves as a patch.
- Second issue was in the parent class `FromArrayProfile`, which still treated the field array as 2D in some cases. This PR implement that the field array is always 3D.
- A third and separate issue is on the way the laser `angularFrequency` is calculated when reading from a non-envelope laser file, such as the ones provided by FBPIC. In these cases, `FromOpenPMDProfile` obtains the central frequency with the `field_to_envelope` method. However, I have observed non-sensical values of this parameter. This affected the vector potential, `a`, to electric field, `E`, envelope transformation when reading Wake-T files, which needs the right value for the central angular frequency of the laser.
- To allow for a proper `a` to `E` transformation when reading from Wake-T envelope files, and to circumvent the problem on the previous point, this PR allows `lambda0` to be passed directly to `FromOpenPMDProfile`, so this value will be used instead of the one calculated by `field_to_envelope`.

### DIFF
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -68,9 +68,7 @@ class FromArrayProfile(Profile):
             )
         else:  # dim = "rt"
             assert axes_order in [["r", "t"], ["t", "r"]]
-            assert len(array.shape) == 3, (
-                "Field array is not 3D."
-            )
+            assert len(array.shape) == 3, "Field array is not 3D."
 
             if axes_order == ["t", "r"]:
                 self.array = np.swapaxes(array, 1, 2)
@@ -81,8 +79,10 @@ class FromArrayProfile(Profile):
             # to make correct interpolation within the first cell
             if axes["r"][0] != 0.0:
                 r = np.concatenate(([-axes["r"][0]], axes["r"]))
-                subarray = array[:,0,:]  # takes first element in second dimension
-                array = np.concatenate((subarray[:, np.newaxis, :], array), axis=1)  # add it at the beginning 
+                subarray = array[:, 0, :]  # takes first element in second dimension
+                array = np.concatenate(
+                    (subarray[:, np.newaxis, :], array), axis=1
+                )  # add it at the beginning
             else:
                 r = axes["r"]
 
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -68,9 +68,12 @@ class FromArrayProfile(Profile):
             )
         else:  # dim = "rt"
             assert axes_order in [["r", "t"], ["t", "r"]]
+            assert len(array.shape) == 3, (
+                "Field array is not 3D."
+            )
 
             if axes_order == ["t", "r"]:
-                self.array = np.swapaxes(array, 0, 2)
+                self.array = np.swapaxes(array, 1, 2)
             else:
                 self.array = array
 
@@ -78,7 +81,8 @@ class FromArrayProfile(Profile):
             # to make correct interpolation within the first cell
             if axes["r"][0] != 0.0:
                 r = np.concatenate(([-axes["r"][0]], axes["r"]))
-                array = np.concatenate(([array[0]], array))
+                subarray = array[:,0,:]  # takes first element in second dimension
+                array = np.concatenate((subarray[:, np.newaxis, :], array), axis=1)  # add it at the beginning 
             else:
                 r = axes["r"]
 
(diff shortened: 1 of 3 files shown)


lasy/profiles/from_array_profile.py
```diff
@@ -68,29 +68,32 @@ def __init__(self, wavelength, pol, array, dim, axes, axes_order=["x", "y", "t"]
             )
         else:  # dim = "rt"
             assert axes_order in [["r", "t"], ["t", "r"]]
+            assert len(array.shape) == 3, "Field array is not 3D."
```
MaxThevenet, line 71: With these changes, can we still read FBPIC output successfully? And use lasy as we do for the FBPIC-Wake-T handshake?

lasy/profiles/from_openpmd_profile.py
```diff
@@ -75,8 +79,10 @@ def __init__(
         field,
         coord=None,
         is_envelope=None,
+        is_waket=None,
```
MaxThevenet, line 82: For the record: we do this at the moment, but before the next release we need to make sure the code work without such a specific option.

lasy/profiles/from_openpmd_profile.py
```diff
@@ -110,13 +116,23 @@ def __init__(
         if not is_envelope:
             grid = create_grid(F, axes, dim, is_envelope=is_envelope)
             grid, omg0 = field_to_envelope(grid, dim, phase_unwrap_nd)
-            array = grid.get_temporal_field()[0]
+            array = grid.get_temporal_field()
         else:
-            s = io.Series(path + "/" + prefix + "_%T.h5", io.Access.read_only)
+            if is_waket:
+                filepath = path + "/" + prefix + "%T.h5"
```
MaxThevenet, line 122: This maybe should be fixed in Wake-T.

lasy/profiles/from_openpmd_profile.py
```diff
@@ -110,13 +116,23 @@ def __init__(
...
             it = s.iterations[iteration]
-            omg0 = it.meshes["laserEnvelope"].get_attribute("angularFrequency")
+            omg0 = it.meshes[field].get_attribute("angularFrequency")
             array = F
 
+        if lambda0:
+            omg0 = 2 * np.pi * c / lambda0
+
+        if field == "a":
+            array = (m_e * c * omg0 / e) * array
```
MaxThevenet, line 134: Could you use `vector_potential_to_field` instead?

MaxThevenet (commented):

MaxThevenet:
Just checking, is this still active?

lasy/profiles/from_array_profile.py
delaossa: Yes

lasy/profiles/from_openpmd_profile.py
delaossa: the `is_waket` flag is only needed to set the right filepath.
For some reason `lasy` assumes a certain file name scheme, e.g. prefix + "_%T.h5".
(see below).

lasy/profiles/from_openpmd_profile.py
delaossa: Here `is_waket` is only used to adapt the filepath. 
A general solution would be to let the user chose the file naming scheme, e.g., by passing directly `filepath` instead of `prefix`.

lasy/profiles/from_openpmd_profile.py
delaossa: Perhaps we could, but what is `vector_potential_to_field` exactly doing here?
```py
A =  -np.gradient(field, grid.axes[-1], axis=-1, edge_order=2) + 1j * omega0 * field
a = m_e * c / e * A
```

delaossa:
There is a new version of this PR, which is meant to be compatible with the new H5 reader here #361.
Check here: https://github.com/delaossa/lasy/tree/read_waket_new
Still work in progress and has no PR yet.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: This is using the definition of vector potential, basically $E_x = -\frac{\partial A_x}{\partial t}$, with correction for the envelope. This makes sure all the terms including CEP are correct.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Yes, this would be great. Either instead of current method (just making it more flexible to accept both), or as an additional input argument.

MaxThevenet:
Superseded by #384.

Issue #345, opened 2025-01-20
Focusing geometry: OAP with hole


alexanatoly:
Hi,

We are focusing the laser pulse using an off axis parabola (OAP) with a hole. Is it possible to simulate such a focusing geometry using lasy ?

Thanks,
Alex

RemiLehe:
Yes, once #357 is merged, this will be possible by combining an intensity mask and a parabola.
(See comments in https://github.com/LASY-org/lasy/pull/357)

alexanatoly:
Hi @RemiLehe and @AnnaLiisaP,

This is an excellent upgrade in lasy. Please, let us know when the ```#357``` is merged. 

Thanks,
Alex

PR #357, opened 2025-02-14
New Optical Element - Intensity Mask

AnnaLiisaP:
Adds functionality to mask out the intensity of different regions of the beam. This could for example be used to mimic the effect of an aperture or a optic with a hole ( see issue #345)

### DIFF
--- a/lasy/optical_elements/intensity_mask.py
+++ b/lasy/optical_elements/intensity_mask.py
@@ -1,6 +1,3 @@
-import numpy as np
-from scipy.constants import c
-
 from lasy.optical_elements.optical_element import OpticalElement
 
 
@@ -20,7 +17,9 @@ class IntensityMask(OpticalElement):
     """
 
     def __init__(self, R, center=(0, 0), aperture_type="aperture"):
-        assert aperture_type in ["aperture", "hole"], "aperture_type must be 'aperture' or 'hole'"
+        assert aperture_type in ["aperture", "hole"], (
+            "aperture_type must be 'aperture' or 'hole'"
+        )
         self.R = R
         self.center = center
         self.aperture_type = aperture_type
@@ -44,7 +43,7 @@ class IntensityMask(OpticalElement):
             Contains the value of the multiplier at the specified points.
             This array has the same shape as the array omega.
         """
-        r_squared = (x - self.center[0])**2 + (y - self.center[1])**2
+        r_squared = (x - self.center[0]) ** 2 + (y - self.center[1]) ** 2
         mask = r_squared <= self.R**2  # True inside, False outside
 
         if self.aperture_type == "aperture":
--- a/lasy/optical_elements/intensity_mask.py
+++ b/lasy/optical_elements/intensity_mask.py
@@ -3,26 +3,31 @@ from lasy.optical_elements.optical_element import OpticalElement
 
 class IntensityMask(OpticalElement):
     """
-    Class for adding an radially symmetric intensity mask that acts as an aperture or a hole.
+    Class for a radially symmetric intensity mask.
+    
+    This creates an optical element which acts to mask out the intensity of a laser pulse. 
+    The mask is radially symmetric and can either mask intensity beyond a user defined radius, thus acting as
+    an aperture. Alternatively, the optica can mask intensity within a user defined radius, acting in this case as 
+    an optic with a hole.
 
     Parameters
     ----------
     R : float (in meter)
-        The Radius of the mask
+        The radius of the mask
     center: tuple (floats)
         Center of the mask. Default is (0,0)
-    aperture_type: string
+    mask_type: string
         Should be 'aperture' (default, allows light inside) or 'hole' (allows light outside).
 
     """
 
-    def __init__(self, R, center=(0, 0), aperture_type="aperture"):
-        assert aperture_type in ["aperture", "hole"], (
-            "aperture_type must be 'aperture' or 'hole'"
+    def __init__(self, R, center=(0, 0), mask_type="aperture"):
+        assert mask_type in ["aperture", "hole"], (
+            "mask_type must be 'aperture' or 'hole'"
         )
         self.R = R
         self.center = center
-        self.aperture_type = aperture_type
+        self.mask_type = mask_type
 
     def amplitude_multiplier(self, x, y, omega):
         """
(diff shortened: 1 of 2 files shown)


lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
+from lasy.optical_elements.optical_element import OpticalElement
+
+
+class IntensityMask(OpticalElement):
+    """
+    Class for adding an radially symmetric intensity mask that acts as an aperture or a hole.
```
rob-shalloo:
```suggestion
    Class for a radially symmetric intensity mask.
    
    This creates an optical element which acts to mask out the intensity of a laser pulse. 
    The mask is radially symmetric and can either mask intensity beyond a user defined radius, thus acting as
    an aperture. Alternatively, the optica can mask intensity within a user defined radius, acting in this case as 
    an optic with a hole.
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Parameters
+    ----------
+    R : float (in meter)
+        The Radius of the mask
```
rob-shalloo:
```suggestion
        The radius of the mask
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+    center: tuple (floats)
+        Center of the mask. Default is (0,0)
+    aperture_type: string
+        Should be 'aperture' (default, allows light inside) or 'hole' (allows light outside).
+
+    """
+
+    def __init__(self, R, center=(0, 0), aperture_type="aperture"):
```
rob-shalloo:
```suggestion
    def __init__(self, R, center=(0, 0), mask_type="aperture"):
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Class for adding an radially symmetric intensity mask that acts as an aperture or a hole.
+
+    Parameters
+    ----------
+    R : float (in meter)
+        The Radius of the mask
+    center: tuple (floats)
+        Center of the mask. Default is (0,0)
+    aperture_type: string
```
rob-shalloo:
```suggestion
    mask_type: string
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+
+    """
+
+    def __init__(self, R, center=(0, 0), aperture_type="aperture"):
+        assert aperture_type in ["aperture", "hole"], (
+            "aperture_type must be 'aperture' or 'hole'"
+        )
+        self.R = R
+        self.center = center
+        self.aperture_type = aperture_type
```
rob-shalloo:
```suggestion
        self.mask_type = mask_type
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns
+        -------
+        multiplier : ndarray of complex numbers
+            Contains the value of the multiplier at the specified points.
+            This array has the same shape as the array omega.
+        """
+        r_squared = (x - self.center[0]) ** 2 + (y - self.center[1]) ** 2
+        mask = r_squared <= self.R**2  # True inside, False outside
+
+        if self.aperture_type == "aperture":
```
rob-shalloo:
```suggestion
        if self.mask_type == "aperture":
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+        The Radius of the mask
+    center: tuple (floats)
+        Center of the mask. Default is (0,0)
+    aperture_type: string
+        Should be 'aperture' (default, allows light inside) or 'hole' (allows light outside).
+
+    """
+
+    def __init__(self, R, center=(0, 0), aperture_type="aperture"):
+        assert aperture_type in ["aperture", "hole"], (
```
rob-shalloo:
```suggestion
        assert mask_type in ["aperture", "hole"], (
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,52 @@
...
+            "aperture_type must be 'aperture' or 'hole'"
```
rob-shalloo:
```suggestion
            "mask_type must be 'aperture' or 'hole'"
```

rob-shalloo (commented):
Hey, thanks for this PR! It's a great new feature!

I would consider adding some more context to the docstring (I've made a suggestion).
Additionally, I think that the `aperture_type` is confusing if one of the types is `aperture` and the other is `hole`, and so I suggest changing the variable name from `aperture_type` to `mask_type`.

lasy/optical_elements/intensity_mask.py
```diff
@@ -0,0 +1,57 @@
...
+        Return the amplitude multiplier.
+
+        Parameters
+        ----------
+        x, y, omega : ndarrays of floats
+            Define points on which to evaluate the multiplier.
+            These arrays need to all have the same shape.
+        omega0 : float (in rad/s)
+            Central angular frequency, as used for the definition
+            of the laser envelope.
```
MaxThevenet: This doesn't seem to be passed as an input, right?
```suggestion
```

MaxThevenet (commented):
Thanks for this PR! See 1 comment below.

MaxThevenet (approved):

MaxThevenet (commented):
The mask is called `intensitymask` but the main function is about the amplitude. Could the naming be consistent? What, in principle, makes most sense? Also, does it make sense to merge intensity and phase masks, as this can be the modulus and phase of a multiplier? Tagging @rob-shalloo too.

rob-shalloo:
Hey, I've just taken another look here. The optic is a mask for the complex amplitude rather than just the amplitude. it simply either sets the complex amplitude to 0 and leaves it as it was. As the intensity is just the absolute value of the complex amplitude squared, the result is equivalent to masking the intensity. In this sense, I think `intensity_mask` is not wrong here and  at least from my perspective is the most intuitive. 

As for optics which modify the intensity versus only modifying the phase. I think we shouldn't differentiate between them. They both operate on the complex amplitude. I also wouldn't see a need to merge them in this case.

MaxThevenet (approved):
Looks great, thanks!

PR #359, opened 2025-02-17
[mini]Fix a bug when weight approaches zero in get_zeta function

huixingjian:
This PR fixed a bug observed when using ```get_zeta``` function  in CI test of HIPACE++ PR#1196.
Sometimes the weight of the average here could be zero, which causes the result of zeta to be NaN. This is now avoided by add an np.where condition to check it.

### DIFF
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -1011,12 +1011,9 @@ def get_zeta(dim, grid, k0):
     weight_y_3d = np.transpose(env_spec_abs2, (2, 0, 1))
     sum_x = np.sum(weight_x_3d, axis=2)
     sum_y = np.sum(weight_y_3d, axis=2)
-    if np.any(sum_x == 0) or np.any(sum_y == 0):
-        print("Warning: Zero weight sum encountered!")
-    xda = np.sum(grid.axes[0] * weight_x_3d, axis=2) / (sum_x + 1e-20)
-    yda = np.sum(grid.axes[1] * weight_y_3d, axis=2) / (
-        sum_y + 1e-20
-    )  # Add a small epsilon
+    eps_wgt = 1e-20 if np.any(sum_x == 0) or np.any(sum_y == 0) else 0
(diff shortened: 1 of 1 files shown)


lasy/utils/laser_utils.py
```diff
@@ -1009,8 +1009,11 @@ def get_zeta(dim, grid, k0):
     # Calculate dx0 and dy0 in (x,y,omega) space
     weight_x_3d = np.transpose(env_spec_abs2, (2, 1, 0))
     weight_y_3d = np.transpose(env_spec_abs2, (2, 0, 1))
-    xda = np.sum(grid.axes[0] * weight_x_3d, axis=2) / np.sum(weight_x_3d, axis=2)
-    yda = np.sum(grid.axes[1] * weight_y_3d, axis=2) / np.sum(weight_y_3d, axis=2)
+    weight_x_2d = np.sum(weight_x_3d, axis=2)
+    weight_y_2d = np.sum(weight_y_3d, axis=2)
+    eps_wgt = 1e-20 if np.any(weight_x_2d == 0) or np.any(weight_y_2d == 0) else 0
+    xda = np.sum(grid.axes[0] * weight_x_3d, axis=2) / (weight_x_2d + eps_wgt)
+    yda = np.sum(grid.axes[1] * weight_y_3d, axis=2) / (weight_y_2d + eps_wgt)
```
MaxThevenet: Not tested, but would sth like this work? Arbitrary values should be avoided when possible.
```suggestion
    xda = np.where(weight_x_2d == 0, 0, np.sum(grid.axes[0] * weight_x_3d, axis=2) / weight_x_2d)
    yda = np.where(weight_y_2d == 0, 0, np.sum(grid.axes[1] * weight_y_3d, axis=2) / weight_y_2d)
```

MaxThevenet (commented):
Thanks! See comment below.

lasy/utils/laser_utils.py
huixingjian: Thanks for the point! this works!

MaxThevenet (approved):

PR #360, opened 2025-02-19
Update polynomial_spectral_phase.py

mjvstreeter:
Added optional delay keyword argument to PolynomialSpectralPhase optical element. Adds capability to delay pulse in time.

### DIFF
--- a/lasy/optical_elements/polynomial_spectral_phase.py
+++ b/lasy/optical_elements/polynomial_spectral_phase.py
@@ -23,6 +23,8 @@ class PolynomialSpectralPhase(OpticalElement):
 
     Parameters
     ----------
+    omega0 : float (in rad/s)
+        Central angular frequency about which the polynomial is expanded
     delay : float (in s), optional
         Group delay (by default: ``delay=0``). Positive value delays the pulse, i.e. it arrives at a later time
     gdd : float (in s^2), optional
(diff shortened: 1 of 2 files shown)


lasy/optical_elements/polynomial_spectral_phase.py
```diff
@@ -33,15 +33,18 @@ class PolynomialSpectralPhase(OpticalElement):
         arriving after the main pulse.
     fod : float (in s^4), optional
         Fourth-order Dispersion (by default: ``fod=0``).
+    delay : float (in s), optional
+        Delay (by default: ``delay=0``). Positive delay moves pulse forward in time.
```
rob-shalloo:
```suggestion
        Group Delay (by default: ``delay=0``). Positive delay moves pulse forward in time.
```

rob-shalloo:
Hey Matt, thanks for the PR! I think this is a great addition and was certainly missing before.

Given that gdd,tod,fod and delay all default to 0, I would suggest reordering the arguments to a more natural order eg: delay, gdd, tod and fod.

Additionally, I would suggest updating the latex equation in the docstring to include the delay term.

I've also added a suggested change to update the name from `Delay` to `Group Delay` in the description of the argument.

lasy/optical_elements/polynomial_spectral_phase.py
mjvstreeter: Thanks for the suggestions. I have incorporated them and also clarified what group delay means.

rob-shalloo:
I've just removed double description of the `delay` parameter and moved `omega0` to first entry in the parameter list as it's the only required argument.

rob-shalloo:
Discussed over zoom, Matt to add test of delay to CI before review

mjvstreeter:
Added the test of delay. 
Also, noticed that grid parameters defined before functions are never used (in case you want to do anything about that).

MaxThevenet (approved):
Thanks for this PR!

Issue #354, opened 2025-02-13
Flattened Gaussian Transverse before and after focusing


rob-shalloo:
The flattened Gaussian transverse profile #349 adds a much needed profile to `lasy`, however, this profile only defines after focusing (ie. near the focus).

For a variety of use cases, it would make sense to have access to both the near field (ie before focusing with an OAP) and far field (i.e near focus) profiles. 

I think we should add the near field profile too, either via an option in the profile or via the creation of a second profile.

PR #363, opened 2025-02-20
Update definition of Flattened Gaussian to include both NF and FF

rob-shalloo:
Updating Alberto's recent addition of a flattened Gaussian beam to include the case where the beam is defined in the nearfield, compared to only defining in the far field. 

Addresses #354

### DIFF
--- a/lasy/profiles/transverse/flattened_gaussian_profile.py
+++ b/lasy/profiles/transverse/flattened_gaussian_profile.py
@@ -109,9 +109,10 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
     def __init__(self, field_type, w, N, wavelength, z_foc=0):
         super().__init__()
         # Ensure that N is an integer
-        self.N = int(round(N))
+        assert isinstance(N, int) and N >= 0
         assert field_type in ["nearfield", "farfield"]
         self.field_type = field_type
+        self.N = N
 
         if field_type == "farfield":
             w0 = w
@@ -185,11 +186,10 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
             w = self.w
 
             sumseries = 0
-            if N > 0:
-                for n in range(N):
-                    sumseries += (
-                        1 / math.factorial(n) * ((N + 1) * (x**2 + y**2) / w**2) ** n
-                    )
+            for n in range(N+1):
+                sumseries += (
+                    1 / math.factorial(n) * ((N + 1) * (x**2 + y**2) / w**2) ** n
+                )
 
             envelope = np.exp(-(N + 1) * (x**2 + y**2) / w**2) * sumseries
 
--- a/lasy/profiles/transverse/flattened_gaussian_profile.py
+++ b/lasy/profiles/transverse/flattened_gaussian_profile.py
@@ -31,11 +31,16 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
         \exp\left(-\frac{r^2}{(N+1)w0^2}\right)
         \sum_{n=0}^N c'_n L^0_n\left(\frac{2\,r^2}{(N+1)w0^2}\right)
 
-    with Laguerre polynomials :math:`L^0_n` and :math:`\qquad c'_n = \sum_{m=n}^{N}\frac{1}{2^m}\binom{m}{n}`
+    
+    with Laguerre polynomials :math:`L^0_n` and 
+    
+    .. math::
+
+        \qquad c'_n=\sum_{m=n}^{N}\frac{1}{2^m}\binom{m}{n}
 
     - For :math:`N=0`, this is a Gaussian profile: :math:`E\propto\exp\left(-\frac{r^2}{w0^2}\right)`.
 
-    - For :math:`N\rightarrow\infty`, this is a Jinc profile: :math:`E\propto \frac{J_1(r/w0)}{r/w0}`.
+    - For :math:`N\rightarrow\infty`, this is a Jinc profile: :math:`E\propto\frac{J_1(r/w0)}{r/w0}`.
 
     The equivalent expression for the collimated beam in the near field which produces this focus is
     given by:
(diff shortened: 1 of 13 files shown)


lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -9,92 +11,126 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
-        The waist of the laser pulse,
-        i.e. :math:`w_{0}` in the above formula.
-    N: int
+    field_type : string
+        Options: 'nearfield', when the beam is defined far from focus and
+        has been collimated, or 'farfield', when the beam is in the vicinity
+        of or has been directly propagated from the focus. In this case there
+        can be a large defocus in the spatial phase.
+
+    w : float (in meter)
```
MaxThevenet, line 75: Is there a reason to change the name `w0`?

lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -9,92 +11,126 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
     ----------
-    w0 : float (in meter)
-        The waist of the laser pulse,
-        i.e. :math:`w_{0}` in the above formula.
-    N: int
+    field_type : string
+        Options: 'nearfield', when the beam is defined far from focus and
+        has been collimated, or 'farfield', when the beam is in the vicinity
+        of or has been directly propagated from the focus. In this case there
+        can be a large defocus in the spatial phase.
```
MaxThevenet: I find this slightly misleading, just tried to help clarify, let me know what you think.
```suggestion
        Options: 'nearfield', when the beam is defined far from focus (e.g., right before the focusing optics), or 'farfield', when the beam is in the vicinity of the focus.
```

lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -9,92 +11,126 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
 
-    - Use a non-zero ``z_foc``
-    - Use ``z_foc=0`` (i.e. initialize the pulse at focus) and then call
-      ``laser.propagate(-z_foc)``
+    In order to initialize the pulse in the far field but out of focus, you
+    must select ``field_type == 'farfield'`` and then you can either:
+
+    - Use a non-zero ``z_foc``.
+    - Use ``z_foc=0`` (i.e., initialize the pulse at focus) and then call
+      ``laser.propagate(-z_foc)``.
```
MaxThevenet: This profile is independent of the propagator, I think this whole warning should be dropped. It says something true, but not required here. I understand this was not introduced in this PR.

lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -9,92 +11,126 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
 
-    The equivalent expression **far from focus** is
+    The equivalent expression for the collimated beam in the near field which produces this focus is
+    given by:
 
     .. math::
 
         E(x,y,z=\infty) \propto
         \exp\left(-\frac{(N+1)r^2}{w(z)^2}\right)
         \sum_{n=0}^N \frac{1}{n!}\left(\frac{(N+1)\,r^2}{w(z)^2}\right)^n
```
MaxThevenet: I find this expression with `w(z)` confusing, as we're talking about a collimated beam. I would call it `R` or `W` or something, and drop the (z) as well as the definition of `w(z)` below, as I do not think it is used when defined in the near field. This would require more changes in the inputs description below.

lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -9,92 +11,126 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
-    - For :math:`N=0`, this is a Gaussian profile: :math:`E\propto\exp\left(-\frac{r^2}{w_(z)^2}\right)`.
+    .. math::
 
-    - For :math:`N\rightarrow\infty`, this is a flat profile: :math:`E\propto \Theta(w(z)-r)`.
+        w(z) = \frac{\lambda_0}{\pi w_0}|z-z_{foc}|
+
+    - Note that a beam defined using the near field definition would be
+      equivalent to a beam defined with the corresponding parameters in
+      the far field, but without the parabolic phase arising from being
+      defined far from the focus.
```
MaxThevenet: Good to have this note here indeed!

MaxThevenet (commented):

lasy/profiles/transverse/flattened_gaussian_profile.py
rob-shalloo: yes, I think having $w_0$ and $w(z)$ in the above formular ensures clarity. It then makes sense to carry this into the definitions below. 
`w0` is used when we're discussing far field and `w` when we're discussing near field

rob-shalloo: I have now removed the w(z) as per comments below, but I still believe `w` and `w0` then are correct

lasy/profiles/transverse/flattened_gaussian_profile.py
rob-shalloo: I agree, I have now modified these to remove the w(z) and to be more consistent with the use of z throughout the docstring

lasy/profiles/transverse/flattened_gaussian_profile.py
rob-shalloo: Agreed, this has been removed

rob-shalloo:
Requested changes implemented

rob-shalloo:
Ready for further review / merge

lasy/profiles/transverse/flattened_gaussian_profile.py
```diff
@@ -42,42 +42,41 @@ class FlattenedGaussianTransverseProfile(TransverseProfile):
...
 
-        E(x,y,z=\infty) \propto
-        \exp\left(-\frac{(N+1)r^2}{w(z)^2}\right)
-        \sum_{n=0}^N \frac{1}{n!}\left(\frac{(N+1)\,r^2}{w(z)^2}\right)^n
+        E(x,y) \propto
+        \exp\left(-\frac{(N+1)r^2}{w^2}\right)
+        \sum_{n=0}^N \frac{1}{n!}\left(\frac{(N+1)\,r^2}{w^2}\right)^n
 
-    with
+    with the relationship between the spot sizes of the beams in the far field and in the near field given by;
```
MaxThevenet:
```suggestion
    with the relationship between the spot sizes of the beams in the far field and in the near field given by:
```

MaxThevenet (approved):

PR #364, opened 2025-02-24
New Optical Element: Zernike Aberrations

rob-shalloo:
Add an optical element to apply Zernike polynomials to a given pulse. 

Currently this makes use of the already existing `zernike.py` utils file. The user provides the pupil coordinates over which the Zernike polynomial is defined and then provides a dictionary with the OSA/ANSI indexing as the keys and zernike amplitudes as the value.


MaxThevenet (approved):
Thanks for this PR!

Issue #358, opened 2025-02-14
Wrong laser wavelength when reading FBPIC output


delaossa:
When reading a laser profile from an FBPIC OpenPMD file, a wrong central laser frequency is extracted by the `field_to_envelope` function.

This code snippet should reproduce the issue:
```py
from lasy.profiles.from_openpmd_profile import FromOpenPMDProfile
from lasy.laser import Laser
from lasy.utils.laser_utils import compute_laser_energy

laser_profile = FromOpenPMDProfile(path='.',
                   iteration=1,
                   pol=(1, 0),
                   field='E',
                   coord='x',
                   theta=0,
                   )
print(f'lambda0 = {laser_profile.lambda0:.3e} m')

dimensions = "rt"
r = laser_profile.axes["r"]
t = laser_profile.axes["t"]
lo = (r.min(), t.min())
hi = (r.max(), t.max())
num_points = (len(r), len(t))
laser = Laser(dimensions, lo, hi, num_points, laser_profile)
energy = compute_laser_energy('rt', laser.grid)
print(f'Energy = {energy :.2f} J')
laser.show()
```

which reads this file (I needed to zip it because github does not allow uploading h5 files):
[fbpic_data.zip](https://github.com/user-attachments/files/18799323/fbpic_data.zip)

This FBPIC data has been obtained with this example -> https://fbpic.github.io/example_input/boosted_frame_script.html

delaossa:
Fix in #376

PR #376, opened 2025-03-21
Fix `FromOpenPMDProfile`

MaxThevenet:
`get_frequency` and `field_to_envelope` broke recently, which caused issue #358. This PR proposes a fix. A slightly more elegant solution could be used if `Grid` could store a Hilbert transform, which is complex but not an envelope, and the current logic assumes is_complex iff is_envelope.

`field_to_envelope` used to return a grid. With this fix, it modifies the input grid in-place, and returns only the central frequency.

Thanks @delaossa for the clear and nice reproducer in #358! On this example, the measured wavelength is now 803.9 nanometers.

### DIFF
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -614,7 +614,7 @@ def field_to_envelope(grid, dim, phase_unwrap_nd=False):
     assert not grid.is_envelope
 
     # Get central wavelength from array
-    omg0_h = get_frequency(
+    _, omg0_h = get_frequency(
         grid,
         dim=dim,
         is_hilbert=False,
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -609,7 +609,7 @@ def field_to_envelope(grid, dim, phase_unwrap_nd=False):
     Returns
     -------
     tuple
-        A tuple with the envelope array and the central wavelength.
+        The central wavelength.
     """
     assert not grid.is_envelope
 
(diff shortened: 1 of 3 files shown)


lucafedeli88:
Hi! 
I've tried to run the reproducer script proposed in https://github.com/LASY-org/lasy/issues/358 before and after the fix. 
In both cases I get the same lambda, but different energies and different profiles.

### Before the fix:
image


###  After the fix: 
image

MaxThevenet:
Oops, thanks @lucafedeli88 for the test! Indeed, I forgot one step in the process, this should be fixed in the last version. The result I get is attached. I think this is the correct one.
Screenshot 2025-03-24 at 12 41 28

What do you call "Before the fix"? Latest dev or latest release? I can reproduce the behavior described by @delaossa.

delaossa:
I have given it a try and it works good, just like @MaxThevenet reported above.

lasy/utils/laser_utils.py
```diff
@@ -609,40 +609,39 @@ def field_to_envelope(grid, dim, phase_unwrap_nd=False):
     Returns
     -------
     tuple
-        A tuple with the envelope array and the central wavelength.
+        The central wavelength.
```
rob-shalloo: Now that you've removed the return of the grid, it should be explictly explained in the docstring that the grid will be updated. Additionally, the descriptor of "Get the compres envelope of a field by applying a Hilbert transform" is no longer correct as really you're getting the central frequency and updating the grid.

Also, here you're returning the `central frequency` not the `central wavelength`

rob-shalloo (commented):
Hey thanks for this PR, it all looks good to me, save one docstring that needs updating. See below.

rob-shalloo (approved):
All looks good! Thanks for the fix!

Issue #377, opened 2025-03-22
Strange/Incorrect values returned by get_phi2()


TimoEichner:
According to its documentation, the [`utils.laser_utils.get_phi2()`](https://github.com/LASY-org/lasy/blob/5a1ccdedf1345383b3340950ba7b1debd68f75ee/lasy/utils/laser_utils.py#L975-L1006) function should return the group delay dispersion (GDD) of the pulse. However, this does not seem to be the case. See for example the following code:

```python
from lasy.laser import Laser
from lasy.profiles import GaussianProfile
from lasy.optical_elements import PolynomialSpectralPhase
from lasy.utils.laser_utils import get_phi2

w0 = 1e-3
tau = 30e-15/np.sqrt(2*np.log(2))
energy = 1
wavelength = 0.8e-6
pol = (1,0)
t_peak = 0

profile = GaussianProfile(wavelength, pol, laser_energy=energy, w0=w0, tau=tau, t_peak=t_peak)

dim='xyt'
lo = (-5e-3, -5e-3, -300e-15)
hi = (5e-3, 5e-3, 300e-15)
npoints = (100, 100, 200)

laser = Laser(dim=dim, lo=lo, hi=hi, npoints=npoints, profile=profile)

gdd_value = 1000e-30

gdd = PolynomialSpectralPhase(laser.profile.omega0, gdd=gdd_value)

laser.apply_optics(gdd)

phi2_output = get_phi2(grid=laser.grid, dim=laser.dim)

print('Calculated:', phi2_output, ', Expected:', gdd_value)
```

which gives the output: `Calculated: (-9.046750219782097e+26, (-5.526846523370059e-28+3.3601099512172264e-27j)) , Expected: 1e-27`.

Looking into the function, it looks like the first of the returned values is the intensity-weighted mean of the second derivative of the temporal phase, rather than the second derivative of the spectral phase (which would be the GDD). 
What the second value is, I haven't yet understood, but it doesn't seem right to me that it is a complex value while the documentation says it should be the GDD in units of `second^2`.

I would suggest to either adapt the documentation to more accurately describe what the function does, or adapt the function to return the correct group delay dispersion.

huixingjian:
[...]

PR #383, opened 2025-03-26
Fix for issue #377: Fixed `grid.get_spectral_field()` and `get_phi2()`, added `get_spectral_phase()` and `get_gdd()` to laser_utils

TimoEichner:
As a proposal for fixing for the problems described in #377 (see also that description for further details):

1. Adapted the conversions between spectral and temporal domain that causes some spectral phase errors in the `grid.get_spectral_field()` function. This could be solved by adding an `fftshift` to the conversions between the spectral and temporal domains.
2. Added a dedicated function that calculates the spectral phase from the `get_spectral_field()` output (similar to `get_spectrum()`)
3. Added a function `get_gdd()` that calculates the group delay dispersion from the second derivative of the spectral phase ($\frac{d^2\varphi(\omega)}{d\omega^2}$).
4. Split up the tasks of `get_phi2()`: Previously it returned both the second derivative of the temporal phase, and the GDD. This was because the temporal phase derivative was required for calculating the GDD. I adapted this to have two dedicated functions for the distinct tasks.
5. Adapted the STC tests that tested for the correct GDD being applied in `GaussianProfile`
6. Fixed an error in the sign of the chirp applied in `GaussianProfile`


huixingjian:
Many thanks for fixing this! I have just some minor comments here.

lasy/utils/grid.py
```diff
@@ -173,9 +173,11 @@ def temporal2spectral_fft(self):
         (Only along the time axis, not along the transverse spatial coordinates.)
         """
         assert self.temporal_field_valid
+
         self.spectral_field = np.fft.ifft(
-            self.temporal_field, axis=time_axis_indx, norm="backward"
+            np.fft.fftshift(self.temporal_field, axes=-1), axis=-1
         )
+
```
huixingjian, line 180: I think it's better if collapse it into two lines to make it easier to read. 
```suggestion
    shifted_temporal = np.fft.fftshift(self.temporal_field, axes=-1) 
    self.spectral_field = np.fft.ifft(shifted_temporal, axis=-1)
```

lasy/utils/grid.py
```diff
@@ -173,9 +173,11 @@ def temporal2spectral_fft(self):
         (Only along the time axis, not along the transverse spatial coordinates.)
         """
         assert self.temporal_field_valid
+
         self.spectral_field = np.fft.ifft(
-            self.temporal_field, axis=time_axis_indx, norm="backward"
+            np.fft.fftshift(self.temporal_field, axes=-1), axis=-1
         )
```
huixingjian: This may look cleaner:
```py
    shifted_temporal = np.fft.fftshift(self.temporal_field, axes=-1) 
    self.spectral_field = np.fft.ifft(shifted_temporal, axis=-1)
```

TimoEichner: Thanks for the comment! I adapted and cleaned up that line, and also the corresponding line in the reverse transformation (`grid.spectral2temporal_fft()`)

lasy/utils/grid.py
```diff
@@ -223,9 +223,10 @@ def temporal2spectral_fft(self):
         (Only along the time axis, not along the transverse spatial coordinates.)
         """
         assert self.temporal_field_valid
-        self.spectral_field = np.fft.ifft(
-            self.temporal_field, axis=time_axis_indx, norm="backward"
-        )
+
+        shifted_temporal = np.fft.fftshift(self.temporal_field, axes=time_axis_indx)
+        self.spectral_field = np.fft.ifft(shifted_temporal, axis=time_axis_indx)
```
MaxThevenet, line 228: I would expect this to change the results for functions using `temporal2spectral_fft`, in particular `laser.apply_optics` and `laser.propagate` but all CI tests pass, I am not sure why at the moment, any idea?

TimoEichner: Good point! After playing around a little and reading on the implementation of [numpy.fft.fft](https://numpy.org/doc/stable/reference/routines.fft.html#implementation-details) and [ifft](https://numpy.org/doc/stable/reference/generated/numpy.fft.ifft.html), I think the (maybe somewhat simplified) explanation is the following:

- The spectral field  that the `laser.apply_optics` and `laser.propagate` methods use, is assumed to be centered on the edges of the array, meaning that zero-frequency component is the first entry of the array (and typically the peak of the envelope is at the edges of the array).

- The way that the DFT is implemented in numpy, both `fft` and `ifft` **always** return an array that has the zero-frequency/zero-time component at the start of the array, independent of the input array. So the output of `grid.temporal2spectral_fft` always satisfies the ordering that `laser.apply_optics` etc expect, independent of whether the additional `fftshift` is in place (as proposed) or not (as it was previously).

- However: Both `np.fft.fft` and `np.fft.ifft` also **expect** an input that has the zero-frequency/zero-time component that the start of the array. The temporal field in LASY (`grid.temporal_field`) is however stored with the zero-time at the center of the array. Because `ifft` does not expect this, it adds an additional phase to the pulse, that corresponds to a shift by half a grid length in the time domain. This additional (wrong) phase seems to be what caused the phase problems described in #377.

- So by adding the `fftshift` we don't change the order of the spectral field after the actual Fourier transform, but  eliminate the additional (wrongly added) phase

lasy/utils/laser_utils.py
```diff
@@ -1285,3 +1283,146 @@ def get_propation_angle(dim, grid, k0):
...
+        - ``'sum'``: Calculates the spectral phase of the spatially summed field (default).
+        - ``'on-axis'``: Calculates the on-axis spectral phase.
+
+    Returns
+    -------
+    phase: ndarray of floats (1D)
+        Spectral phase of the pulse in the specified units and calculation method (in rad/s)
+
+    omega: ndarray of floats (1D)
+        Angular freuqencies at which the phase is defined (in rad/s)
```
```suggestion
        Angular frequencies at which the phase is defined (in rad/s)
```

lasy/utils/laser_utils.py
```diff
@@ -1285,3 +1283,146 @@ def get_propation_angle(dim, grid, k0):
...
+    """
+    # Field must be envelope
+    assert grid.is_envelope
+
+    # get the spectral field
+    field_spectral = grid.get_spectral_field()
+    field_spectral = np.fft.fftshift(field_spectral, axes=-1)
+
+    # if method=='on-axis' get the on-axis field envelope, and calculate its phase
+    if method == "on-axis":
```
MaxThevenet, line 1346:
```suggestion
    assert method in ['on-axis', 'sum']
    if method == "on-axis":
```

lasy/utils/laser_utils.py
```diff
@@ -1285,3 +1283,146 @@ def get_propation_angle(dim, grid, k0):
...
+    phase = np.unwrap(phase)
+
+    # create omega array (angular frequencies)
+    omega = np.fft.fftshift(
+        2 * np.pi * np.fft.fftfreq(grid.npoints[-1], grid.dx[-1]), axes=-1
+    )
+
+    if omega0:
+        omega += omega0
+
```
MaxThevenet: `omega0` isn't optional, is it? I think this should always be done.
```suggestion
    omega += omega0

```

lasy/utils/laser_utils.py
```diff
@@ -1285,3 +1283,146 @@ def get_propation_angle(dim, grid, k0):
...
+    -------
+    gdd: ndarray of floats (1D)
+        Group delay dispersion over the entire spectral range (in s^2)
+
+    gdd0: float
+        Group delay dispersion at the center frequency (in s^2)
+
+    """
+    # calculate the spectral phase of the laser pulse
+    phase, omega = get_spectral_phase(grid, dim, method=method, omega0=omega0)
```
MaxThevenet, line 1423: Strictly speaking, I think the frequency used in the definition of the laser envelope (see [readme](https://github.com/LASY-org/lasy)) can differ from the central frequency and from the frequency at which the GDD is evaluated. Would it be more accurate to let `get_gdd` take 2 arguments:
 - `omega0`, the frequency at which the envelope is defined and that is passed to `get_spectral_phase`
 - `omega_gdd`, the frequency at which the use choses to evaluate the GDD.

What do you think?

MaxThevenet (commented):
Looks great, thanks for this PR! See some comments attached. I think we can merge first, then I'll update #390.

lasy/utils/laser_utils.py
TimoEichner: You are completely right. I added this to the function, but kept `omega_gdd` optional, with `omega0` being the default frequency at which the GDD is calculated. In my opinion this keeps the usage a little more simple and more similar to some of the other `laser_utils`. What do you think?

lasy/utils/laser_utils.py
```diff
@@ -1399,7 +1399,10 @@ def get_gdd(grid, dim, omega0, method="sum"):
                     Cylindrical (r) transversely, and temporal (t) longitudinally.
 
     omega0 : float
-        Central angular frequency at which the GDD is calculated
+        Central angular frequency of the field
+
+    omega_gdd : float, optional
+        Central angular frequency at which the GDD is calculated, if `None` the central frequency `omega0` is used.
```
```suggestion
        Angular frequency at which the laser envelope is defined.

    omega_gdd : float, optional
        Central angular frequency at which the GDD is calculated, if `None`, `omega0` is used.
```

lasy/utils/laser_utils.py
```diff
@@ -1423,6 +1426,10 @@ def get_gdd(grid, dim, omega0, method="sum"):
...
     gdd = np.gradient(gd, omega, axis=-1)
 
-    # get the GDD at the center frequency
-    gdd0 = np.interp(omega0, omega, gdd)
+    # get the GDD at the specified frequency...
+    if omega_gdd:
+        gdd0 = np.interp(omega_gdd, omega, gdd)
+    # ... or at the central frequency
+    else:
+        gdd0 = np.interp(omega0, omega, gdd)
```
```suggestion
    # get the GDD at the specified frequency or the envelope's frequency
    omega_eval = omega_gdd if omega_gdd is not None else omega0
    gdd0 = np.interp(omega_eval, omega, gdd)
```

lasy/utils/laser_utils.py
MaxThevenet: Agreed, that's better, thanks!

MaxThevenet (approved):
Looks great, thanks! See minor final suggestions below.

PR #361, opened 2025-02-19
[WIP] Modified h5 reader for LASY

Paaaaarth:
This pull request includes several changes to the `lasy` package, focusing on improving the handling of laser profiles from array and OpenPMD files, as well as utility functions for laser manipulation. The most important changes include modifications to the `FromArrayProfile` and `FromOpenPMDProfile` classes, and updates to utility functions in `laser_utils.py`.

### Changes to `FromArrayProfile` class:
* Corrected the axis swapping logic in the `__init__` method to ensure the array is properly oriented.
* Enhanced the handling of the radial axis to ensure correct interpolation within the first cell.

### Changes to `FromOpenPMDProfile` class:
* Refactored the class to use the `openpmd_api` instead of `OpenPMDTimeSeries`, removing unnecessary parameters and adding new ones for better functionality. 
* Improved the reading and processing of OpenPMD files, including handling different field dimensions and converting vector potential to electric field when necessary.

### Updates to utility functions:
* Modified the `create_grid` function to use `np.allclose` for axis comparison, ensuring better numerical stability.
* Added a new `chunk_to_slice` function to convert `openPMD_api.ChunkInfo` to slices, facilitating chunk-based data reading.This pull request includes significant updates to the `lasy` package, focusing on improving the handling of laser profiles from arrays and OpenPMD files. Key changes include fixing axis swapping in the `FromArrayProfile` class, updating the `FromOpenPMDProfile` class to use the `openpmd-api` for reading data, and enhancing the `create_grid` function to handle different dimensions accurately.

**Available functionality:**

- [x] Reads LASY files (With single mode)
- [x] Reads Wake-T files
- [ ] Reads FBPIC files (https://github.com/Paaaaarth/LASY/pull/3)

PR #384, opened 2025-04-04
Fix and generalize FromOpenPMDProfile

MaxThevenet:
This PR proposes a generalisation of profile FromOpenPMDProfile to remove the dependency on [openPMD-viewer](https://github.com/openPMD/openPMD-viewer) and use directly [openpmd-api](https://github.com/openPMD/openPMD-api), and clarify the handling of azimuthal modes to also allow multiple modes. Built on #361 by @Paaaaarth and @delaossa.

# Main changes in the code
- dependency on `openpmd-viewer` was dropped.
- openpmd-api should now be `>=0.16.1`.
- `FromArrayProfile.evaluate` was limited to mode 0, now it allows for higher modes.
- `FromArrayProfile` is less permissive and only takes in arrays with the right axes ordering `x,y,t` and `r,t` (in the past, it could also take `t,y,x` and `t,r`). This is because `FromArrayProfile` is a low-level base class that should remain simple, to be used either internally or by experienced users.
- `FromOpenPMDProfile` was largely re-written (mostly by @Paaaaarth & @delaossa), its input arguments are changed etc.

# Satellite changes
- File `utils/openpmd_output.py` was renamed `utils/openpmd_helper.py` and contains new functions useful for reading openPMD data.
- In `laser.py`, when initialising a Laser object in `rt` geometry, we use `profile.evaluate(x,y,t)` and sample a few points in the azimuthal direction. A new function was added `evaluate_rt(mode,r,t)` was added, such that going from a cylindrical profile to a cylindrical laser does not go through a Cartesian representation. Currently this function is only defined for `FromArrayProfile`, but this could be generalised.

# Tests
### Done, See [this notebook](https://github.com/user-attachments/files/19685094/openpmd.reader.pdf)
- FBPIC -> RT
- FBPIC -> XYT
- LASY XYT -> XYT
- LASY XYT -> RT
- LASY RT -> RT
- Wake-T -> RT
- HiPACE++ -> XYT
- HiPACE++ -> RT

### To do
- Check with multiple modes, e.g. create an FBPIC profile where all the data is in mode 3, and see what it gives.


lasy/profiles/from_array_profile.py
```diff
@@ -49,65 +52,72 @@ def __init__(self, wavelength, pol, array, dim, axes, axes_order=["x", "y", "t"]
         super().__init__(wavelength, pol)
 
         assert dim in ["xyt", "rt"]
+        assert array.ndim == 3, "array myst be 3D, [x,y,t] or [modes,r,t]."
```
MaxThevenet:
```suggestion
        assert array.ndim == 3, "array must be 3D, [x,y,t] or [modes,r,t]."
```

lasy/profiles/from_array_profile.py
```diff
@@ -49,65 +52,72 @@ def __init__(self, wavelength, pol, array, dim, axes, axes_order=["x", "y", "t"]
...
 
-            # The `RegularGridInterpolator` below expects a 2D array.
-            # However, when reading lasy envelope files, the array is 3D.
-            # First dimension corresponds to the azimuthal mode decomposition.
-            # For now, this only fix profiles with one mode.
-            if len(array.shape) == 3:
-                assert array.shape[0] == 1, (
-                    "Handling `rt` profiles with more than one azimuthal mode still needs to be implemented."
+            self.field_interp_modes = []
+            # Loop over the 2*m-1 elements of the array and createe a separate
```
MaxThevenet:
```suggestion
            # Loop over the 2*m-1 elements of the array and create a separate
```

lasy/profiles/from_openpmd_profile.py
```diff
@@ -13,116 +16,79 @@ class FromOpenPMDProfile(FromArrayProfile):
     r"""
     Profile defined from an openPMD file.
 
+    Upon initialization, read from an openPMD profile, builds interpolation objects on the array data and use them to create function evaluate.
```
MaxThevenet:
```suggestion
    Upon initialization, read from an openPMD profile, build interpolation objects on the array data and use them to create function evaluate.
```

lasy/laser.py
```diff
@@ -131,25 +131,34 @@ def __init__(
...
+                y = r * np.sin(theta)
+                # Evaluate the profile on the generated grid
+                envelope = profile.evaluate(x, y, t)
+                # Perform the azimuthal decomposition
+                azimuthal_modes = np.fft.ifft(envelope, axis=0)
+                field = azimuthal_modes[:n_azimuthal_modes]
+                if n_azimuthal_modes > 1:
+                    field = np.concatenate(
+                        (field, azimuthal_modes[-n_azimuthal_modes + 1 :])
+                    )
```
MaxThevenet, line 161: This was unchanged.

lasy/utils/openpmd_helper.py
```diff
@@ -0,0 +1,303 @@
...
+import numpy as np
+import openpmd_api as io
+from scipy.constants import c
+
+from lasy import __version__ as lasy_version
+
+from .laser_utils import field_to_vector_potential
+
+
+def write_to_openpmd_file(
```
MaxThevenet, line 12: This function was not changed in this PR.

delaossa:
Hi @MaxThevenet, thanks a lot for this! It looks great and also worked perfectly with my test scripts.

I only have a couple of comments for now regarding the use of the `FromOpenPMDProfile` class:
1. Perhaps `filename` should be renamed to `filepath`?
2. Passing `iteration` as argument of `FromOpenPMDProfile` would be interesting when using a file pattern instead of a specific file for `filename`, e.g. `"data%T.h5"`. I think that this will be useful when analyzing the laser from a simulation in a loop or when selecting which iteration to restart a simulation from. (This could go in a separate PR).
3. I wonder if there could be a default behavior for `is_envelope`. For example, we could let `is_envelope = False` by default and make it `True` automatically when `field_name` is not `None`. That would simplify a little the use of `FromOpenPMDProfile`, e.g.:
- For the full field, we just pass `filename` (the code would automatically get the right combination of the electric field from file, anyways):
```
laser_profile = FromOpenPMDProfile(filename='data00000001.h5')
```
- For the envelope, we give the `field_name` (which is necessary in this case):
```
laser_profile = FromOpenPMDProfile(filename='data00000001.h5', field_name="laserEnvelope")
```
(Minor: I just notice that the naming of `filename` and `field_name` are not consistent to each other.)

5. Not completely sure if this logic works well:
```py
   try:
       envelopeField = m.get_attribute("envelopeField")
       pol = m.get_attribute("polarization")
   except Exception:
       envelopeField = "normalized_vector_potential"
       pol = (1, 0)
```
What if `envelopeField` attribute is present, but `polarization` is not? Wouldn't `envelopeField` become `"normalized_vector_potential"` in all cases?

MaxThevenet:
Thanks for your comments!
1. I think filepath could be misunderstood as only the path, which is what openPMD-viewer could take, so I wanted to avoid confusion with this name.
2. For the sake of simplicity, I would rather avoid the additional machinery in there, the current version should be tolerant enough. But we can reassess if this is really restrictive if e.g. 1 openPMD file contains multiple time steps.
3. You're right, that's better, I implemented it.
4. filename was renamed file_name
5. You are correct with the behavior description. All codes should have both attributes, so they are dealt with together.

The notebook attached above was update correspondingly.

lasy/profiles/from_openpmd_profile.py
```diff
@@ -13,116 +16,73 @@ class FromOpenPMDProfile(FromArrayProfile):
...
+                m = i.meshes[field]
+                component = coord_list[count]
+                axes_order, axes, array = extract_array(m, series, component)
+                array_list.append(array)
+            array = convert_modes(array_list, geometry, is_envelope, verbose)
+            dim = "xyt" if geometry == "cartesian" else "rt"
+            grid = create_grid(array, axes, dim, is_envelope=False)
+            omg0 = field_to_envelope(grid, dim)
+            array = grid.get_temporal_field()
+            pol = (1, 0)
```
rob-shalloo, line 85: Does this mean the polarisation of the laser is hardcoded in? What if the openpmd file contains a y-polarized pulse? I'm not so familiar with the inner workings of openpmd, but from a first glance at the code its not obvious how this is accommodated as `coord_list` only has `x` in cartesian. If this is the case, it should be detailed in the docstring.

delaossa (approved):
It is working great with the examples that I have! Including the FBPIC -> Wake-T handshake.
Thanks!

rob-shalloo:
Thanks very much for this PR! As is, I think this could be merged, I've got one comment (see above) which may require a doc string modification / explanation and the only other major comment I would have is that in certain places we've started using the 3 letters to describe the three dimensions of the array in the azimuthal decomposition, eg `mrt` axes ordering, for example in the `FromArrayProfile` where there is an assert:
```
assert array.ndim == 3, "array must be 3D, [x,y,t] or [modes,r,t]."
```
yet in most of the code we revert to `rt` instead of `mrt`, for example in the same class when we define the axes ordering.
```
assert axes_order == ["r", "t"]
```

Does it make sense to make this more uniform in this PR? Given that it is anyway a non-backward compatible change? Of course it might also make sense to wait and do a full overhaul at a later point.

lasy/profiles/from_openpmd_profile.py
MaxThevenet: Thanks. Indeed, we currently assume that the input from an electromagnetic openPMD file represents a x-polarized laser pulse. Two options: let the user define it or (even better) detect it in the output. This would be suitable for a follow-up PR IMO. I will update the docstring accordingly.

MaxThevenet:
Thanks for your review! For the user, we kept the `rt`/`xyt` at the moment, although we use `mrt` in a naming internally. I agree this would warrant discussion and maybe we could make it more consistent, but you're right it should be separate from this PR.

rob-shalloo:
Ok then agreed we move the `mrt` topic to another discussion / future PR.

Everything else looks great, thanks for updating the docstring.

Thanks in general again for a great PR!

rob-shalloo (approved):

Issue #352, opened 2025-02-13
Spectrum flipped?


MaxThevenet:
There is a suspicion that, in some cases, the spectrum read is flipped, at least with some plotting functions. @Paaaaarth and @delaossa can provide more information.

MaxThevenet:
@Paaaaarth @delaossa could you put a reproducer here?

delaossa:
Hello,
Here a reproducer:

```py
import numpy as np
import scipy.constants as ct
from lasy.laser import Laser
from lasy.profiles.combined_profile import CombinedLongitudinalTransverseProfile
from lasy.profiles.longitudinal import LongitudinalProfileFromData
from lasy.profiles.transverse.gaussian_profile import GaussianTransverseProfile
from lasy.utils.laser_utils import get_spectrum
import matplotlib.pyplot as plt

# Parameters
pol = (1, 0)  # Linearly polarized in the x direction
laser_energy = 25  # Pulse energy in Joules

# Longitudinal profile
file_longitudinal = 'Spectrum.txt'
exp_wavelength = np.loadtxt(file_longitudinal, usecols=0, dtype="float")  # nanometers
exp_wavelength = exp_wavelength * 1e-9 # meters
exp_spectrum = np.loadtxt(file_longitudinal, usecols=1, dtype="float")  # Arbitary units
longitudinal_data = {
    "datatype": "spectral",
    "axis_is_wavelength": True,
    "axis": exp_wavelength,
    "intensity": exp_spectrum,
    "dt": 1e-15,
}
tmax = 500e-15
longitudinal_profile = LongitudinalProfileFromData(longitudinal_data, lo=-tmax, hi=tmax)
# print(f'lambda0 = {longitudinal_profile.lambda0*1e6:.3f} um')

# Transverse profile
waist = 50e-6
transverse_profile = GaussianTransverseProfile(waist, longitudinal_profile.lambda0)

# Combined profile
laser_profile = CombinedLongitudinalTransverseProfile(
    wavelength=longitudinal_profile.lambda0,
    pol=pol,
    laser_energy=laser_energy,
    long_profile=longitudinal_profile,
    trans_profile=transverse_profile,
)

# Create Laser object from profile
dimensions = "rt"  # Use cylindrical geometry
t = longitudinal_profile.time,
lo = (0, np.min(t))
hi = (3 * waist, np.max(t))
num_points = (500, 1000)
laser = Laser(dimensions, lo, hi, num_points, laser_profile)

[...]

PR #388, opened 2025-04-08
[mini] Fix issue with flipped spectrum

rob-shalloo:
@delaossa and @Paaaaarth identified an issue with the `LongitudinalProfileFromData` class, in which when an asymmetric spectrum was passed as an input, the retrieved spectrum from the `Laser` object was flipped (see #352 ).

After looking into this, it seems to arise from different conventions for the fourier transform from spectral to temporal space. `LongitudinalProfileFromData` moves from spectral to temporal space with an inverse fourier transform, while in the `Grid` and indeed in the `get_spectrum` function, this process involves a fourier transform.

The fix is to replace the inverse fourier transform in `LongitudinalProfileFromData` with a fourier transform (similarly the `fftshift and `ifftshift` have to be replaced).


MaxThevenet (approved):
Thanks for the fix!

PR #390, opened 2025-04-09
[almost mini] More consistant handling of omega axis

MaxThevenet:
`Grid.get_spectral_field` returns the field in `x,y,omega` space where the `x` and `y` axis are available in `Grid`'s metadata but not `omega`. This PR proposes to let `get_spectral_field` return the `omega` axis in addition to the spectral field, and use this consistently everywhere instead of recalculating it. In addition, this PR proposes a similar handling of `fftshift` in `get_spectrum` and `get_spectral_phase`.


lasy/utils/grid.py
```diff
@@ -226,6 +232,7 @@ def temporal2spectral_fft(self):
         self.spectral_field = np.fft.ifft(
             self.temporal_field, axis=time_axis_indx, norm="backward"
         )
+        self.spectral_axis = 2 * np.pi * np.fft.fftfreq(self.npoints[-1], self.dx[-1])
```
rob-shalloo, line 234: Does this need to be recalculated here? If nothing else in the grid changes, then surely the original one calculated when the grid was created would work?

MaxThevenet: Good point; in the case one changes the properties of an existing `grid` object (not sure if it's done anywhere currently), this would be safer. This operation is cheap so it should be alright to do it every time.

lasy/utils/grid.py
```diff
@@ -122,6 +122,7 @@ def __init__(
         self.temporal_field_valid = False
         self.spectral_field = np.zeros(self.shape, dtype="complex128")
         self.spectral_field_valid = False
+        self.spectral_axis = 2 * np.pi * np.fft.fftfreq(npoints[-1], self.dx[-1])
```
MaxThevenet:
```suggestion
```

Issue #396, opened 2025-04-17
Color map in laser.show()


CriMariani:
Laser.show() works by default, but when I try to change colormap I get this error

```py
TypeError                                 Traceback (most recent call last)
Cell In[88], line 1
----> 1 laser.show(show_intensity=True, cmap =ffwd_cmap)

File ~/.conda/envs/icsenv/lib/python3.9/site-packages/lasy/laser.py:438, in Laser.show(self, show_intensity, **kw)
    425 def show(self, show_intensity=False, **kw):
    426     """
    427     Show a 2D image of the laser amplitude or intensity.
    428 
   (...)
    436     **kw : additional arguments to be passed to matplotlib's imshow command
    437     """
--> 438     show_laser(self.grid, self.dim, show_intensity, **kw)

File ~/.conda/envs/icsenv/lib/python3.9/site-packages/lasy/utils/plotting.py:96, in show_laser(grid, dim, show_intensity, **kw)
     94 divider = make_axes_locatable(ax)
     95 cax = divider.append_axes("right", size="5%", pad=0.05)
---> 96 im = ax.imshow(
     97     F_plot, extent=extent, cmap="Reds", aspect="auto", origin="lower", **kw
     98 )
     99 cb = fig.colorbar(im, cax=cax)
    100 cb.set_label(cbar_label)

TypeError: matplotlib.axes._axes.Axes.imshow() got multiple values for keyword argument 'cmap'
```

reproducer using the Gaussian tutorial
```py
from lasy.laser import Laser
from lasy.profiles.gaussian_profile import GaussianProfile

wavelength = 800e-9  # Laser wavelength in meters
polarization = (1, 0)  # Linearly polarized in the x direction
energy = 1.5  # Energy of the laser pulse in joules
spot_size = 25e-6  # Waist of the laser pulse in meters
pulse_duration = 30e-15  # Pulse duration of the laser in seconds
t_peak = 0.0  # Location of the peak of the laser pulse in time

laser_profile = GaussianProfile(
    wavelength, polarization, energy, spot_size, pulse_duration, t_peak
)

dimensions = "rt"  # Use cylindrical geometry
lo = (0, -2.5 * pulse_duration)  # Lower bounds of the simulation box
hi = (5 * spot_size, 2.5 * pulse_duration)  # Upper bounds of the simulation box
num_points = (300, 500)  # Number of points in each dimension

laser = Laser(dimensions, lo, hi, num_points, laser_profile)

[...]

PR #401, opened 2025-04-22
Fixed laser.show() colormap

em-archer:
Proposed fix for #396

Fixed bug with calling alternative colormaps in laser.show()


MaxThevenet (approved):

PR #405, opened 2025-04-23
Fix bug in openPMD reader when extracting central frequency

MaxThevenet:
In the `FromOpenPMDProfile` init in cylindrical geometry, the envelopes **and** frequencies of Ex and Ey was extracted separately, and the latest value calculated (for Ey) was used when setting up the final LASY profile. For an x-polarized laser pulse, this contained garbaged causing the frequency to be incorrect (and probably sometimes the polarization). This PR proposes a fix by determining which field contains most signal, getting the central frequency from there and using it in the definition of the envelope of the other field.

Thanks @delaossa for reporting and documenting (offline)!


rob-shalloo (approved):
Hi, thanks for reporting @delaossa and for fixing @MaxThevenet ! Looks good to me.

PR #408, opened 2025-04-28
[WIP] Implement a Fresnel Propagator 

rob-shalloo:


MaxThevenet:
Thanks for this PR! I think we should merge this soon into `props` so we really have a working example. Here are some general comments:
- Style: could you replace names like `postFactor` with `postfactor`? Could you improve the name for variable `F` with something more specific? If possible, keep all snake case. For the meshgrids, could you also avoid caps? You could consider `mg_` to specify it's a meshgrid.
- The idea is that `Propagator.propagate` is the only place where `update` is called, and where the actual initialization is done. In the future (but not yet), we'll have some logic in `update` such that the propagator is initialised only if needed. For instance, if we propagate multiple times on the same grid for the same distance d, the propagator will be initialised only in the first `propagate` call. Could you check whether you could put more things in your `update`?
- Let's turn off the interpolation for the moment, to get the machinery working albeit slightly less accurate.
- Currently `Propagator.propagate` takes `grid_in` and returns the propagated `grid_out`. I think this could be useful e.g. to do some checks in `Laser.propagate`, but we could also modify in place, let me know.
- Could you fix documentation and docstrings etc. so CI passes, except for tests using the propagator? For the latter, we'll use the same axiprop propagator as before, to avoid changing too many things together.

PR #411, opened 2025-04-28
Structure for new propagators

MaxThevenet:
This PR proposes a structure for a new class `Propagator` and derived classes to allow for the implementation of new propagators in addition to current Axiprop propagators. Previously, once a `Laser` object was initialised, the propagation was done with
```py
laser.propagate(distance)
```
With the current implementation, this would still be supported (using a default propagator, currently Axiprop) but the user could also choose a propagator by doing
```py
from lasy.propagators import FresnelSFFTPropagator
propagator = FresnelSFFTPropagator()
laser.add_propagator(propagator)
laser.propagate(distance)
```

# Design choices
- Each propagator derives from base class `Propagator`, and implements `propagate` and `update` member functions.
- The lines above should work for both dimensions, such that `dim` is handled internally with the Propagator class.
- The distance propagated is stored in `grid.position` instead of shifting time axis (see #418).
- A propagator should implement a `propagate` function that also takes care of updating the propagator if needed.

# Main changes
- A centralised `fft` function was created in `utils/fft_wrapper.py` that should be used for all FFTs done within the LASY codebase. Two reasons: make it easier to port to GPU with a centralised approach, and make sure we use the same convention everywhere for fft/ifft, fftshifts, etc. (these conventions are detailed in the docstring).
- The Axiprop propagators were moved to `propagators/axiprop_propagators.py` (implemented by @hightower8083). A skeleton of the code reads as below (there are 2 such propagators in `axiprop_propagators.py`). Only function `propagate` should be used from outside the class.
```py
class AxipropPropagator(Propagator):
    def update() # call _update_mrt or _update_xyt
    def _update_mrt()
    def _update_xyt()
    def propagate() # call update and call _propagate_mrt or _propagate_xyt
    def _propagate_mrt()
    def _propagate_xyt()
```


rob-shalloo:
I would suggest the following modification to calling the propagator
```
z_R = 3.14159 * spot_size**2 / wavelength

# old method
# laser.propagate(-z_R)  # Propagate the pulse upstream of the focal plane

# new method
from lasy.propagators import FresnelSFFTPropagator
propagator = FresnelSFFTPropagator()
laser.add_propagator(propagator)
laser.propagate(-z_R)
```

ie. the last step I've changed from `laser.TMP_propagate(-z_R)` to `laser.propagate(-z_R)`. We should be clever enough to correctly use the propagator without further specifying it after `add_propagator`. This also means changing the propagator is only one line of code rather than 2.

Incidentally, I don't think we should store more than one propagator at a time.

lasy/propagators/axiprop_propagators.py
```diff
@@ -0,0 +1,641 @@
...
+
+        omega0 : float (in rad.s^-1)
+            The main frequency :math:`\omega_0`, which is defined by the laser
+            wavelength :math:`\lambda_0`, as :math:`\omega_0 = 2\pi c/\lambda_0`.
+
+        containers_in : Axiprop container(s)
+            An Axiprop container (dim='xyt'), or list of containers (dim='rt', 1 element per mode), with the data of laser to propagate.
+
+        grid_out : Grid object (optional)
+            Grid object for on which the propagated laser pulse is defined.
```
TimoEichner: Minor typo
```suggestion
            Grid object on which the propagated laser pulse is defined.
```

lasy/propagators/axiprop_propagators.py
```diff
@@ -0,0 +1,641 @@
...
+        ----------
+        omega0 : float (in s^-1)
+            The main frequency :math:`\omega_0`, which is defined by the laser
+            wavelength :math:`\lambda_0`, as :math:`\omega_0 = 2\pi c/\lambda_0`.
+
+        containers_in : Axiprop container(s)
+            A list of Axiprop containers, with the data of laser to propagate.
+
+        grid_out : Grid object (optional)
+            Grid object for on which the propagated laser pulse is defined.
```
TimoEichner: Minor typo, same as above. This seems to also be present in documentation of `grid_out` in other methods in this file

TimoEichner (commented):
Thanks for this structure and nice base for adding propagators! In general it looks good and usable to me - only a couple of very minor typos. 

One comment/question came up while going through the PR: probably it would make sense to have something similar to `fft_wrapper.py` also for `rt` geometry to allow for easier implementation of `rt` in various upcoming propagators. Will that be part of a separate PR?

MaxThevenet:
Thanks for the comments! I typos fixed. As for xyt vs rt, good point, this could be kept as is replacing `if transverse` with `if transverse and dim == 'xyt'`, or split in separate functions. But indeed, let's differ the implementation or `rt` for a future PR, as this will require Hankel transforms with a bit more work.

PR #413, opened 2025-04-28
Show laser stuff

delaossa:
Some extra features for `show_laser`:

- One can specify the spatial and temporal units for the axes by passing a dictionary.
- Time axis can be shifted automatically or by the given amount.
- Option to plot the normalized vector potential.

laser_profile


lasy/laser.py
```diff
@@ -422,17 +422,52 @@ def write_to_file(
...
+        field_type : string, default: "field"
+            Options are:
+            - ``'field'``: Show the envelope of the laser field.
+            - ``'intensity'``: Show the intensity of the laser field.
+            - ``'vector_potential'``: Show the vector potential of the laser field.
+
+        t_shift : float, default: 0
+            Shift the temporal axis by `t_shift` seconds.
+            It also can be a string with `"left"`, `"right"` or `"center"`,
+            to shift the temporal axis to the left, right or center of the time axis.
```
rob-shalloo: I think this functionality is excellent, but this description should probably be a bit clearer. My understanding is that in the case of a string, the function will shift the temporal axis such that the t=0 lies at the left, right or center of the x-axis.

lasy/laser.py
```diff
@@ -422,17 +422,52 @@ def write_to_file(
...
 
         Parameters
         ----------
-        show_intensity : bool
-            if False the laser amplitude is plotted
-            if True then the intensity of the laser is plotted along with lineouts
-            and a measure of the pulse duration and spot size
+        field_type : string, default: "field"
+            Options are:
+            - ``'field'``: Show the envelope of the laser field.
```
rob-shalloo, line 323: I would suggest the use of `envelope` or `electric_field_enevelope`. Otherwise a `fieldtype` of `field` is two different uses and meanings of the word field

lasy/utils/plotting.py
```diff
@@ -3,11 +3,21 @@
...
-
-def show_laser(grid, dim, show_intensity, **kw):
-    """
+from .laser_utils import field_to_vector_potential, get_duration, get_w0
+
+
+def show_laser(
+    grid,
+    dim,
+    field_type="field",
```
rob-shalloo: same mention of `field` as in `laser.py`

lasy/utils/plotting.py
```diff
@@ -23,10 +33,31 @@ def show_laser(grid, dim, show_intensity, **kw):
...
+    field_type : string, default: "field"
+        Options are:
+        - ``'field'``: Show the envelope of the laser field.
+        - ``'intensity'``: Show the intensity of the laser field.
+        - ``'vector_potential'``: Show the vector potential of the laser field.
+
+    t_shift : float, default: 0
+        Shift the temporal axis by `t_shift` seconds.
+        It also can be a string with `"left"`, `"right"` or `"center"`,
+        to shift the temporal axis to the left, right or center of the time axis.
```
rob-shalloo: Same comment as in `laser.py`

rob-shalloo (commented):
Hey @delaossa , thanks very much for this PR. I only have two very small comments related to the doc string, but otherwise I think it looks great!

One final thing which would be great to add prior to merging is an documentation page for the plotting utility. This would mean adding a `.rst` file to the `docs` folder

lasy/laser.py
delaossa: That is exactly what it does. I have used your sentence for clarity in a new commit.

lasy/laser.py
delaossa: I have changed it to `envelope_type`.

delaossa:
Address Rob's comments and add `propagation` to the plot's labels.

delaossa:
> One final thing which would be great to add prior to merging is an documentation page for the plotting utility. This would mean adding a `.rst` file to the `docs` folder

Oh I forgot this. I'll take a look..

rob-shalloo (approved):
Hi @delaossa , thanks again for the PR and for making the proposed changes. This functionality is a great addition to lasy!

MaxThevenet (approved):
Thanks for this PR!

PR #416, opened 2025-04-29
[WIP] Props axiprop

hightower8083:
a follow-up PR for #411 that adds
- wrappers to axiprop propagators 
-  method to add_propagator to the `laser` object (temporal, for tests)


MaxThevenet:
Thanks for this PR! I think we should merge this soon into `props` so we really have a working example. Note that I modified the code a bit to help consistency between `props` and @rob-shalloo's Fresnel propagator. Here are some general comments:
- Could the LASY propagator classes be e.g. `AxipropFresnelPropagator` and `Propagator` and handle the dim internally? It can even be in different files, etc., but it would be good to not have to change the propagator when changing the dim in a lasy script, that's only a user aspect.
- The idea is that `Propagator.propagate` is the only place where `update` is called, and where the actual initialization is done, where you recommended it should be. In the future (but not yet), we'll have some logic in `update` such that the propagator is initialised only if needed. For instance, if we propagate multiple times on the same grid for the same distance d, the propagator will be initialised only in the first `propagate` call. Could you check whether you could put more things in your `update`? From what we discussed offline, I think this could be the case.
- Currently `Propagator.propagate` takes `grid_in` and returns the propagated `grid_out`. I think this could be useful e.g. to do some checks in `Laser.propagate`, but we could also modify in place, let me know.
- Could you use the appropriate Axiprop version in CI such that the tests pass with the new framework? Then we can merge in dev, that will be very useful. Maybe something like `pip install -U --no-deps git+https://github.com/hightower8083/axiprop.git@main` or so, not sure.
- Could you fix documentation and docstrings etc. so CI passes?

PR #418, opened 2025-05-05
[mini] Store position separately

MaxThevenet:
Currently, the grid object contains a time axis that contains both the comoving coordinate for the longitudinal profile of the laser pulse and the position at which the laser pulse is defined. Hence, if the laser is defined on a grid `[tmin, tmax]`, propagating by distance `d` means we shift the grid to `[tmin+d/c, tmax+d/c]`. This is entirely correct, but it requires a bit of handling in the propagators, and makes it slightly less obvious when propagating in a medium with group velocity `vg<c`. This small PR proposes to split these two elements, and keep `[tmin, tmax]` centred around `0` while encoding the distance propagated in a separate attribute `grid.position`, also written to file. The user is responsible for defining the origin of this position axis (as before). ~~I think this could be added to the openPMD standard for laser envelope.~~ in openPMD, this is encoded in `grid_global_offset`.

### DIFF
--- a/lasy/utils/openpmd_helper.py
+++ b/lasy/utils/openpmd_helper.py
@@ -88,6 +88,7 @@ def write_to_openpmd_file(
     # Store metadata needed to reconstruct the field
     m.set_attribute("angularFrequency", 2 * np.pi * c / wavelength)
     m.set_attribute("polarization", pol)
+    m.set_attribute("position", grid.position)
     if save_as_vector_potential:
         m.set_attribute("envelopeField", "normalized_vector_potential")
         m.unit_dimension = {}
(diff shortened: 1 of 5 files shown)


lasy/utils/openpmd_helper.py
```diff
@@ -88,6 +88,7 @@ def write_to_openpmd_file(
     # Store metadata needed to reconstruct the field
     m.set_attribute("angularFrequency", 2 * np.pi * c / wavelength)
     m.set_attribute("polarization", pol)
+    m.set_attribute("position", grid.position)
```
RemiLehe: Instead of storing this as a separate attribute, could modify how we save `m.grid_global_offset` to include this shift? This would be more consistent with the openPMD standard and would avoid having to impose changes on the way other codes read this quantity.

RemiLehe (commented):
Thanks for this PR!
I have a minor suggestion on the openPMD saving.

RemiLehe (approved):

PR #421, opened 2025-05-09
Implement `AngularSpectrumDFFTPropagator` and `NonlinearKerrStep`

TimoEichner:
Added several components to allow propagation in dispersive media and intensity dependent phase-modulation due to Kerr nonlinearity. When combined in a split-step approach, these allow to simulate self-focusing as well as changes to the spectral pulse properties due to self-phase modulation. The key parts are the following:

### 1. `AngularSpectrumDFFTPropagator`:
A double FFT propagator based on the angular spectrum propagation method. This allows straightforward simulation in dispersive media by considering the (wavelength dependent) refractive index of the medium. The refractive index is passed to the propagator when initializing it or through the `update()` method as the parameters `n`. It is 1 by default to describe propagation in vacuum, but can also be an array of the refractive index along the frequency/wavelength axis or the grid that is passed to the `propagate()` method, or a callable function of wavelength.

Tests are included (in `test_angular_spectrum_propagator.py`) to ensure that the propagator output matches analytical formulas for a) the caustic of a Gaussian beam around a focus , and b) the temporal broadening of a Gaussian pulse due to group velocity dispersion.

### 2. `NonlinearKerrStep`:
Has a very similar structure to a propagator, but actually only applies an intensity dependent phase in the `xyt` domain and does not propagate the pulse. This is implemented in `propagators/nonlinear_phase_shift.py` to also allow possible future nonlinear effects and to have them all in the same file. Whether it makes sense to leave them in the `propagators` directory can be discussed, but in my opinion this would make sense as the nonlinear phase shifts would typically be used together with propagation, and have a very similar structure. 
A key difference is replacing  the `.propagate()` method by `.apply()`, but otherwise usage should be very similar to the `Propagator`s.

A test is included (`test_spm.py`) that ensures that the spectral broadening of a Gaussian pulse due to self-phase modulation matches the expected value from an analytical equation for Gaussian pulses (from the Agrawal book on nonlinear fiber optics). 

To analyze the bandwidth, I also added a function `get_bandwidth()` to `utils/laser_utils.py`. Useage is very similar to `get_duration()`, except for additional optional arguments to get the bandwidth not as the standard deviation, but at a given intensity level (e.g. FWHM), to get the bandwidth either on-axis or spatially integrated, and to get the bandwidth either as angular frequency or in units of wavelength.

### 3. A tutorial on using these two classes and on how to combine them to simulate nonlinear propagation in media.
Example usage of both of the above described classes is shown in the tutorial, by going through the process of simulating the propagation of an intense laser pulse through a fused silica window. The pulse sees dispersion of the medium, as well as spectral broadening and self-focusing due to the Kerr nonlinearity.


lasy/propagators/angular_spectrum_dfft_propagator.py
```diff
@@ -0,0 +1,189 @@
...
+
+    dim : string
+        Dimensionality of the array. Options are:
+
+        - ``'xyt'``: The laser pulse is represented on a 3D grid:
+                    Cartesian (x,y) transversely, and temporal (t) longitudinally.
+        - ``'rt'`` : The laser pulse is represented on a 2D grid:
+                    Cylindrical (r) transversely, and temporal (t) longitudinally.
+
+    n : int, float, 1d array or callable, Optional
```
MaxThevenet:
```suggestion
    n : int, float, 1d array or callable, optional
```

lasy/propagators/angular_spectrum_dfft_propagator.py
```diff
@@ -0,0 +1,189 @@
...
+        assert isinstance(n, (int, float, np.ndarray)) or callable(n)
+        assert dim in ["rt", "xyt"]
+
+        self.n = n  # refractive index
+        self.omega0 = omega0
+        self.dim = dim
+
+    def propagate(self, distance, grid_in, dim=None, omega0=None, grid_out=None):
+        r"""
+        Propagates the laser field in z diration by a given distance using the angular spectrum method.
```
MaxThevenet:
```suggestion
        Propagates the laser field in z direction by a given distance using the angular spectrum method.
```

lasy/propagators/angular_spectrum_dfft_propagator.py
```diff
@@ -0,0 +1,189 @@
...
+            wavelength :math:`\lambda_0`, as :math:`\omega_0 = 2\pi c/\lambda_0`.
+
+        grid_out : Grid object (optional)
+            Grid object on which the propagated laser pulse is defined.
+            Can be different from laser grid before propagation.
+
+        Returns
+        -------
+        Grid object with laser data after propagation.
+
```
MaxThevenet:
```suggestion
```

lasy/utils/laser_utils.py
```diff
@@ -1455,3 +1455,115 @@ def get_dispersion(grid, dim, omega0, order, omega_eval=None, method="sum"):
...
+    float
+        Spectral bandwidth of the pulse in the specified units and calculation method.
+    """
+    # Get the volume of each grid cell and spectral field
+    dV = get_grid_cell_volume(grid, dim)
+    field, omega = grid.get_spectral_field()
+
+    # Calculate omega
+    #   omega = np.fft.fftfreq(grid.npoints[-1], grid.dx[-1]) * 2 * np.pi
+
```
```suggestion
```

lasy/utils/laser_utils.py
```diff
@@ -1455,3 +1455,115 @@ def get_dispersion(grid, dim, omega0, order, omega_eval=None, method="sum"):
...
+    # Choose axis along which to calculate the bandwidth
+    if unit == "m":  # convert omega to wavelength
+        assert omega0, "'omega0' must be provided to calculate bandwidth in meters."
+        width_axis = 2 * np.pi * c / (omega + omega0)
+
```
MaxThevenet: Just keeping if-else together without newline, as done below.
```suggestion
```

lasy/utils/laser_utils.py
```diff
@@ -1455,3 +1455,115 @@ def get_dispersion(grid, dim, omega0, order, omega_eval=None, method="sum"):
...
+    else:  # keep omega as that axis
+        width_axis = omega
+
+    # Calculate weights of each grid cell (amplitude of the field).
+    if dim == "xyt":
+        spectral_intensity = np.abs(field) ** 2 * dV
+
```
```suggestion
```

docs/source/tutorials/nonlinear_propagation_split_step.ipynb
```diff
@@ -0,0 +1,372 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "51833efa",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "linear_proapgator = AngularSpectrumDFFTPropagator(\n",
```
MaxThevenet: Could you fix typo in `linear_proapgator` here and below?

lasy/propagators/angular_spectrum_dfft_propagator.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        dim = dim if dim is not None else self.dim
+        omega0 = omega0 if omega0 is not None else self.omega0
+        self.update(omega0=omega0, dim=dim, n=self.n)
```
MaxThevenet: Could the checks be done within `self.update`?
```suggestion
        self.update(omega0=omega0, dim=dim, n=self.n)
```
The idea of `update` is described in the corresponding base class's docstring:
> Typically called at the beginning of the propagate function, to update the propagator itself if required.
If the update is costly, some machinery should be in place to do it only if required.
Some propagators need to be updated if the distance to propagate changes.
Others only need to be updated if the grid properties change, etc.

In your case, I don't think there's expensive operations to do, so `update` is not doing much but that's fine.

I am also not sure about the logic within `update`: would this be safer?
```py
self.dim = dim if dim is not None else self.dim
assert self.dim is not None
```

MaxThevenet (commented):
Thanks for this PR! Really cool features and tutorial!! Here are general comments, I think we can merge soon after these are addressed.
- In the end, with the fft wrapper, it is not clear that there will be much common operation between DFFT propagators. Could you rename the main class to `AngularSpectrumPropagator`? Still good to mention in docstring that it is of type double FFT, but there will not be a `class DoubleFFTPropagator(Propagator)` anytime soon I think.
- For angular spectrum and Kerr stuff, could you add a corresponding file in https://github.com/LASY-org/lasy/tree/development/docs/source/api/propagators? You will find plenty of examples there, this will put the corresponding doc in the API section of readthedocs.
- In your current implementation, I think we lose the information on the exact timing (the pulse is kept around t=0 even when vg != c). Could we preserve this information by shifting the time axis correspondingly? Not sure how to compute this exactly, @hightower8083 can comment.
- In tutorials and in test, you're using `AngularSpectrumPropagator.propagate()`. Could you use `Laser.propagate` instead? This should be simpler and clearer. If that cover all your cases, you could consider making `dims` and `omega0` mandatory parameters of `AngularSpectrumPropagator.propagate()`. The user would rarely interact with this function directly, so having a few more arguments is OK.

In addition, see minor comments below.

MaxThevenet (approved):
Looks great, thanks!

PR #423, opened 2025-05-09
fix argument order in axiprop fresnel propagator

hightower8083:
Currently the Axiprop Fresnel wrapper doesn't work due to the different order of arguments defined in the propagate method. This PR fixes it


MaxThevenet (approved):
Oops, thanks for the fix!

PR #427, opened 2025-05-20
[mini] make split step tutorial second order in time

MaxThevenet:
Suggested by @RemiLehe the split step tutorial can be made second order in time.


docs/source/tutorials/nonlinear_propagation_split_step.ipynb
```diff
@@ -218,11 +218,12 @@
     "\n",
     "for i in range(Nsteps):\n",
     "    # linear propatation step\n",
```
MaxThevenet:
```suggestion
    "    # linear propagation 1/2 step\n",
```

docs/source/tutorials/nonlinear_propagation_split_step.ipynb
```diff
@@ -218,11 +218,12 @@
...
-    "    laser.propagate(dz)\n",
+    "    laser.propagate(dz / 2)\n",
     "\n",
     "    # nonlinear propagation step\n",
     "    laser.grid = nonlinear_phase.apply(grid_in=laser.grid, distance=dz)\n",
     "\n",
+    "    laser.propagate(dz / 2)\n",
```
MaxThevenet, line 227:
```suggestion
    "    # linear propagation 1/2 step\n",
    "    laser.propagate(dz / 2)\n",
```

hightower8083:
hey @MaxThevenet,

The PR looks good to me! 
I couldn't resist to check the agreement between the 1st and 2nd order schemes, so I've added the error calculation as following
```py
Nsteps = 10  # number of propagation steps
dz = d_FS / Nsteps  # length of the individual propagation steps

laser1 = deepcopy(laser)
linear_propagator1 = AngularSpectrumPropagator(
    omega0=profile.omega0, n=n_fusedsilica, dim=laser.dim
)

laser.add_propagator(linear_propagator)
laser1.add_propagator(linear_propagator1)

err_rel = []

for i in range(Nsteps):
    # propagate first order
    laser1.propagate(dz)
    laser1.grid = nonlinear_phase.apply(grid_in=laser1.grid, distance=dz)

    # propagate second order
    laser.propagate(dz / 2)
    laser.grid = nonlinear_phase.apply(grid_in=laser.grid, distance=dz)
    laser.propagate(dz / 2)

    err_abs = np.abs(laser.grid.get_temporal_field()-laser1.grid.get_temporal_field()).sum()
    err_norm = max(np.abs(laser.grid.get_temporal_field()).sum(), np.abs(laser1.grid.get_temporal_field()).sum())
    if err_norm>0:
        err_rel.append( err_abs / err_norm ) 
    else:
        err_rel.append( 0.0 ) 

    # print progress
    print(f"Progress: {100 * (i + 1) / Nsteps:.1f}%", end="\r")
```

The discrepancy between the methods grows linearly with propagation and for 10 steps reaches ~5%.  
Figure 2(6)

As expected, using smaller steps allows to reduce this error, but it keeps growing in any case. Note, that this error defines the  rather the accuracy of the previous (1st order method) and not the new one, so we may only conclude that 1st order method may have some stability issues.

With this error estimate, I've tried adding the adaptive step size as following
```py
z_loc = 0.0
while z_loc<d_FS:
    if dz > L_sim - z_loc:
        dz = L_sim - z_loc

    [.....]
    z_loc += dz
    ErrFact = err_max / err_rel[-1]
    dz *= 0.9 * ErrFact**(1./2)
```

This worked, but is rather sensitive to parameters of step correction -- the formula with error factor is adjusted with `0.9` multiplier, which is a free parameter and may affect convergence.

So i guess we can merge this tuto as it is, but in future consider some check with a know analytic solution, otherwise having the step-size as a free parameter without any error control may turn be risky.

hightower8083 (approved):

MaxThevenet:
Thanks for the test! Yes we are planning a testing phase for these solvers.

PR #439, opened 2025-06-23
Fresnel Propagator with ZoomFFT

rob-shalloo:
Implementation of a Fresnel Propagator which utilises the Chirp-Z Transform (or Zoom FFT) 

Normally, the Fourier transform is computed using the Fast Fourier Transform (FFT) algorithm.

However, in this case, the Chirp-Z Transform (or Zoom FFT) is used to compute the Fourier transform. This allows for more flexibility in choosing both the initial and final sampling of the Fourier transform.

The algorithm is based upon the work by Hu et al., https://www.nature.com/articles/s41377-020-00362-z and the implementation of the Chirp-Z Transform in SciPy, specifically `scipy.signal.zoom_fft`.


lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -0,0 +1,266 @@
...
+
+from .propagator import Propagator
+
+
+class FresnelChirpZPropagator(Propagator):
+    r"""Class that represents a Fresnel propagator based upon the Chirp-Z Transform.
+
+    The propagated field is calculated via the following method:
+
+    Given a scalar field :math:`E_0(x',y',0,\omega)`, one write the propagated field
```
MaxThevenet:
```suggestion
    Given a scalar field :math:`E_0(x',y',0,\omega)`, one writes the propagated field
```

MaxThevenet (commented):
LGTM, thanks for this PR! See one typo, then should we merge? @NadezhdaKHACHAT you could give it a try, it may help in your case.

rob-shalloo:
Sounds great! Let's Merge :D

MaxThevenet (approved):

Issue #446, opened 2025-07-31
Sampling frequency in Fresnel Propagator with ZoomFFT


rob-shalloo:
Sampling frequency in the `_zoomFourierTransform2D` method is calculated as
```
sample_frequency_x = len(x) / x_range
```

This should be 
```
sample_frequency_x = (len(x)-1) / x_range
```
and similarly for the y-axis

Otherwise introduces a phase tilt.

Issue #444, opened 2025-07-23
Fresnel Propagator with ZoomFFT does not raise an error when passing dim='rt'


NadezhdaKHACHAT:
New Fresnel Propagator with ZoomFFT (https://github.com/LASY-org/lasy/pull/439) does not work in rt coordinates, however it is possible to pass dim='rt' for it

PR #448, opened 2025-07-31
[mini] Small fixes to ChirpZ Propagator

rob-shalloo:
Addresses issues #444  and #446 

Additionally, modified to correctly carry beam position through calculations


MaxThevenet (approved):
Thanks for this PR!

PR #452, opened 2025-08-08
Rectangle

lasycora:
Updated the intensity_mask optical element, to allow for rectangular apertures.

A new string input parameter allows to distinguish between rectangular and round apertures, the default is round.

If the aperture is rectangular, the feature R (radius, if the aperture is round) can also be passed as a tuple, that specifies the halfwidth and the halfheight. If only a single value is given and the aperture is rectangular, a quadratic shape is assumed.

Only the definition of the mask is differs for rectangular in comparison to round apertures.

### DIFF
--- a/lasy/optical_elements/intensity_mask.py
+++ b/lasy/optical_elements/intensity_mask.py
@@ -3,11 +3,11 @@ from lasy.optical_elements.optical_element import OpticalElement
 
 class IntensityMask(OpticalElement):
     """
-    Class for a radially symmetric intensity mask.
+    Class for a round or rectangular intensity mask.
 
     This creates an optical element which acts to mask out the intensity of a laser pulse.
-    The mask is radially symmetric and can either mask intensity beyond a user defined radius, thus acting as
-    an aperture. Alternatively, the optica can mask intensity within a user defined radius, acting in this case as
+    The mask is either round or rectangular and can either mask intensity beyond a user defined radius or width, thus acting as
+    an aperture. Alternatively, the optica can mask intensity within a user defined radius or width, acting in this case as
     an optic with a hole.
 
     Parameters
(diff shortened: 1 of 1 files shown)


lasy/optical_elements/intensity_mask.py
```diff
@@ -3,31 +3,43 @@
...
     """
-    Class for a radially symmetric intensity mask.
+    Class for a round or rectangular intensity mask.
 
     This creates an optical element which acts to mask out the intensity of a laser pulse.
-    The mask is radially symmetric and can either mask intensity beyond a user defined radius, thus acting as
-    an aperture. Alternatively, the optica can mask intensity within a user defined radius, acting in this case as
+    The mask is either round or rectangular and can either mask intensity beyond a user defined radius or width, thus acting as
+    an aperture. Alternatively, the optica can mask intensity within a user defined radius or width, acting in this case as
     an optic with a hole.
```
MaxThevenet, line 11:
```suggestion
    The mask is either round or rectangular, and can either mask intensity outside of the mask (aperture) or inside the mask (hole).
```

lasy/optical_elements/intensity_mask.py
```diff
@@ -3,31 +3,43 @@
...
 
     Parameters
     ----------
-    R : float (in meter)
-        The radius of the mask
+    R : float (in meter) or tuple (floats)
+        The radius of the mask for round masks or half-width and half-height for rectangular masks as tuple.
+        If the shape is rectangular and only one number is given, a quadratic shape is assumed.
```
MaxThevenet, line 17:
```suggestion
    R : float (in meter) or tuple (floats)
        For a round mask, 1 value for the radius of the mask.
        For a rectangular mask, 2 values for the half size in x and in y, respectively. If only 1 value is given, the mask is assumed to be square.
```

MaxThevenet (changes requested):
Thanks for this PR! See minor suggestions for docstrings. Also, from automated tests [results](https://github.com/LASY-org/lasy/actions/runs/16900408158/job/47879057276?pr=452), there seems to be a mistake in the code, a missing parenthesis or something. Could you have a look and confirm with a test that this does what we want? This test needs not to be automated, but you could run something on your laptop and share some figures in the PR description.

PR #466, opened 2025-10-27
Nonlinear propagation tutorial: Added soliton propagation example

TimoEichner:
To properly compare the nonlinear split step propagation against an analytical example, I extended the tutorial by an example of soliton pulse propagation in optical fibers, which can be analytically described in a 1D case sech^2-shaped pulses. 
An example of such propagation of a 1550nm pulse in a fused silica fiber was added to the tutorial. The pulse is initiated with the Soliton peak power at which the pulse shape and power should remain constant during propagation due to a balance of dispersion and nonlinearity, and this constant propagation was verified by letting the pulse propagate over long distances. The power and duration of the pulse indeed stay constant to within the permille level (I would attribute the minor difference to higher order dispersion which is included in the model, but is neglected by the analytic description).

To support the improvements of the tutorial, I also added a few helpers at other locations:
- ```Sech2Profile``` is similar to the other longitudinal pulse profiles but describes a squared hyperbolic secant intensity profile that is required for soliton pulse propagation.
- ```get_duration()``` was extended to be able to calculate not only the rms pulse duration, but also the duration at specific intensity levels such as the FWHM. This is set by the ```level``` parameter of the function. To calculate the width at the specified level, also a new function ```get_width_at_level()``` was added to ```laser_utils.py```
- ```AngularSpectrumPropagator```was updated to handle 1d pulse profiles.


lasy/propagators/angular_spectrum_propagator.py
```diff
@@ -115,18 +116,26 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
+        elif (grid_in.shape[0] == 1) and (grid_in.shape[1] == 1):
+            field = self._propagate_1d(distance, grid_in)
 
         else:  # self.dim == "xyt"
-            field, dt = self._propagate_xyt(distance, grid_in)
+            field = self._propagate_xyt(distance, grid_in)
+
+        omega = frequency_axis("longitudinal", grid_in.axes[-1], "real")
+
+        field, dt = self._compensate_group_delay(field, distance, omega)
```
MaxThevenet: Could you add an input parameter in `propagate` to let the user control whether the group delay compensation is controlled (by default True, so same behaviour as now)? I know this is not a new feature of this PR, but this could help in some cases.

MaxThevenet (commented):
Looks great, thanks for this PR! See 1 minor comment below. Also, could you fix the merge conflicts? Sorry for the delay...

Issue #467, opened 2025-11-15
Phase artifacts after creating laser from experimental data using ```FromArrayProfile```


TimoEichner:
Under some circumstances, there seems to be a problem with the way the field -- and in particular the phase/wavefront -- is interpolated when creating a laser using experimental data and ```FromArrayProfile```. This seems to be due to the way that the field is interpolated from the arrays axes/grid to the new grid of the ```Laser```instance, which is using a single interpolator for the whole complex envelope. This seems to cause problems in the interpolation of the phase, but on the amplitude and only be a problem if the grid of the array and the grid of the new ```Laser```are different.

Below are the wavefront and amplitude for three different cases to show the problem:

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/4be562d5-68f2-4071-b27f-43f8b0d1f07b" />

#### Case1

Laser created using exactly the same grid as in the experimental data

```
file = np.load('data_file.npz')

t = file['time']
x = file['x']
y = file['y']

dimensions = "xyt" 

# Grid is exactly the same as in the file
lo = (x.min(),y.min(),t.min()) 
hi = (x.max(),y.max(),t.max())  
num_points = (x.size, y.size, t.size)

profile = FromArrayProfile(wavelength= file['center_wavelength'],
                           pol=(0,1), 
                           array=file['E_xyt'],
                           dim=dimensions,
                           axes={'x':x,
                                 'y':y,
                                 't':t},
                           axes_order=['x', 'y', 't'])

laser = Laser(dimensions, lo, hi, num_points, profile)
```

#### Case 2

A new custom grid is made and the field is interpolated to that field by ```FromArrayProfile```

```
file = np.load('data_file.npz')

t = file['time']
x = file['x']
y = file['y']

dimensions = "xyt" 

# Custom grid is provided
lo = (-300e-6,-300e-6,-500e-15)
hi = (300e-6,300e-6,500e-15)
num_points = (200, 200, 300)


profile = FromArrayProfile(wavelength= file['center_wavelength'],
                           pol=(0,1), 
                           array=file['E_xyt'],
                           dim=dimensions,
                           axes={'x':x,
                                 'y':y,
                                 't':t},
                           axes_order=['x', 'y', 't'])

laser = Laser(dimensions, lo, hi, num_points, profile)
```

#### Case 3

[...]

PR #469, opened 2025-11-18
Fixed interpolation ```FromArrayProfile``` to solve issue #467

TimoEichner:
Adapted the method of interpolation in the 'xyt' field, where previously a single ```RegularGridInterpolator``` was used to interpolate the whole complex field envelope, to now have two interpolators: one for the real part and one for the imaginary part. This fixes the problems described in #467


MaxThevenet (approved):
Thanks for this PR! When I saw the date when you opened the PR, it was a bit embarrassing to ask you to fix the conflicts, so I just did it and pushed to your branch directly, hope it's OK.

AlexanderSinn:
@MaxThevenet @TimoEichner This PR breaks a HiPACE++ CI test where an xyt pulse is read in and analyzed by LASY (https://github.com/Hi-PACE/hipace/actions/runs/24927811825/job/73000638115?pr=1375).
To me it looks like evaluate expects the abs and phase to be in the real and imag part of combined_field and then the line `xp.abs(xp.real(combined_field)) * xp.exp(1.0j * xp.imag(combined_field))` converts it back. But with the PR, combined_field is already in a real + imag format, but it gets converted anyway.

PR #473, opened 2025-12-16
Fix for some issues in `from_array_profile`

hightower8083:
Recently we've used lasy to pass laser from FBPIC to WARPX and I've noticed a few new issues that made our workflow crash:
- after passing to GPU support [this assertion](https://github.com/LASY-org/lasy/blob/d1bac094ff5b73f0756e9965716d38fb67be19aa/lasy/utils/laser_utils.py#L873) started giving False with differences of ~1e-300. Using `xp.all` with `==` condition on floats is in general risky, so I've changed it to a proper `xp.allclose`
- the use of `xp.concatinate` with GPU data [here](https://github.com/LASY-org/lasy/blob/d1bac094ff5b73f0756e9965716d38fb67be19aa/lasy/profiles/from_array_profile.py#L75) gives error as it needs strictly array structures, while numpy work with lists
- for interpolation [here](https://github.com/LASY-org/lasy/blob/d1bac094ff5b73f0756e9965716d38fb67be19aa/lasy/profiles/from_array_profile.py#L97) we use phase 1D unwrapping along `t`-axis which however leave phase along `r` quiet piecewise. Changing the unwrapping direction fixes the issue.
 
For the latter issue, this practically messes up the laser as following (need a resolved `get_full_field` data to see):
Figure 1(11)

and the fixed version looks as it should:
Figure 2(14)

### DIFF
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -73,7 +73,6 @@ class FromArrayProfile(Profile):
             if axes["r"][0] != 0.0:
                 # add mirrored point to the axis
                 r = xp.concatenate((-axes["r"][[0]], axes["r"]))
-                #r = xp.concatenate((xp.array([-axes["r"][0]]), axes["r"]))
                 # takes first element of the array in the radial dimension
                 subarray = self.array[:, 0, :]
                 # add it at the beginning to be the value at the mirrored point
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -73,7 +73,7 @@ class FromArrayProfile(Profile):
             if axes["r"][0] != 0.0:
                 # add mirrored point to the axis
                 r = xp.concatenate((-axes["r"][[0]], axes["r"]))
-                #r = xp.concatenate((xp.array([-axes["r"][0]]), axes["r"]))
+                # r = xp.concatenate((xp.array([-axes["r"][0]]), axes["r"]))
                 # takes first element of the array in the radial dimension
                 subarray = self.array[:, 0, :]
                 # add it at the beginning to be the value at the mirrored point
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -72,7 +72,8 @@ class FromArrayProfile(Profile):
             # to make correct interpolation within the first cell
             if axes["r"][0] != 0.0:
                 # add mirrored point to the axis
-                r = xp.concatenate(([-axes["r"][0]], axes["r"]))
+                r = xp.concatenate((-axes["r"][[0]], axes["r"]))
+                #r = xp.concatenate((xp.array([-axes["r"][0]]), axes["r"]))
                 # takes first element of the array in the radial dimension
                 subarray = self.array[:, 0, :]
                 # add it at the beginning to be the value at the mirrored point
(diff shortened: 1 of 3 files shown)


AlexanderSinn:
Thanks for these fixes! Do you know if the current dev version still works when using the NumPy backend and if this PR works for both backends? In particular, I'm wondering about the axis change in Unwrap. Does the unwrap axis in `if dim == "xyt":` also need to be changed?

hightower8083:
hi @AlexanderSinn,

The first two points do not appear with numpy backend -- the roundup errors in `xp.all` with `==` appear only on GPUs  and `numpy.concatinate` accepts `array_like` while `cupy.concatinate` only takes arrays.

For the phase unwrapping, typically the temporal phase dynamics is well resolved and varies less along propagation (at given $r_0$), while in radial directions phase changes faster and gets even faster with focusing, which more likely to produce unwrapping artifacts that mess up the interpolations.   For `xyt` I would also prioritize unwrapping transverse plane, e.g. using double `xp.unwrap(xp.unwrap(..., axis=1), axis=0)` instead of `xp.unwrap(..., axis=-1)` [here](https://github.com/LASY-org/lasy/blob/d1bac094ff5b73f0756e9965716d38fb67be19aa/lasy/profiles/from_array_profile.py#L64), but this has to be checked with data. One risk is that 1D unwrapping runs from first to last element, which for `mrt` starts with the on-axis field, where phase behaves well, and in `xyt` it'll start at the boundaries where the  phase of nearly-zero field may run kinda crazy. I played with [such stuff in axiprop](https://github.com/hightower8083/axiprop/blob/62f54c10f36c0eaf19284b81b5729b29be0ed166/axiprop/utils.py#L484), but it wasn't conclusive.

lasy/profiles/from_array_profile.py
```diff
@@ -72,7 +72,8 @@ def __init__(self, wavelength, pol, array, dim, axes, axes_order=["x", "y", "t"]
             # to make correct interpolation within the first cell
             if axes["r"][0] != 0.0:
                 # add mirrored point to the axis
-                r = xp.concatenate(([-axes["r"][0]], axes["r"]))
+                r = xp.concatenate((-axes["r"][[0]], axes["r"]))
+                # r = xp.concatenate((xp.array([-axes["r"][0]]), axes["r"]))
```
AlexanderSinn: OK thanks, this makes sense. Can you remove the commented-out line

AlexanderSinn (approved):

hightower8083:
Hey @AlexanderSinn and @MaxThevenet one more fix is needed. 

In our workflow with files and profiles I've noticed another bug [here](https://github.com/LASY-org/lasy/blob/d1bac094ff5b73f0756e9965716d38fb67be19aa/lasy/utils/openpmd_helper.py#L90). For the cases with `wavelength` calculated from non-enveloped field, the expression `2 * xp.pi * c / wavelength` appears as a cupy structure and openpmd-api fails to set the attribute, so we have to replace it with `to_cpu(2 * xp.pi * c / wavelength)`.

I could add this fix in this PR, though its not concerns the profile directly, or do another PR. Let me know what you prefer.

cheers and happy end of the year holidays!

AlexanderSinn:
You can add it to this PR. I had this issue a lot with CuPy, where it returns zero-dimensional arrays instead of values. The fix I used is to add `float(...)`, not sure if `to_cpu(...)` works in this case.

hightower8083:
the issue is again the tendency of cupy to always assume array type -- `cupy.average()` returns a 0-dim array, while `numpy.average` makes a float. I'm not fully sure about recasting `cupy.array` with a `float()` -- it works but gives a `DeprecationWarning`. From my tests, use of  `to_cpu` fixes the issue since these arrays are handled by `openpmd-api`'s `set_attribute` method correctly. I'll check it once again and will add to this PR

AlexanderSinn:
@hightower8083 is this PR ready to merge?

hightower8083:
@AlexanderSinn, yes, its fine with me -- I've done a few more tests and didn't have any issues with `to_cpu` so its all good, and does  fixes both phase unwrap and openpmd writer bugs

AlexanderSinn (approved):
I tested the LASY CI on GPU and all checks passed on this branch. Note that the code changed by this PR likely isn't tested, or else the issue would have been caught before.

PR #369, opened 2025-03-05
Refractiveindex

pots007:
This PR adds the capability to calculate the the refractive index of a variety materials using the refractiveindex.info database.  

Some features:
- The database is downloaded upon first use, then used from there on
- Material definition follows the `book`, `shelf`, `page` convention from refractiveindex.com. Some common materials can be initialised by just their name.
- For any material, the refractive index can be calculated.
- The spectral phase expansion terms dphi/dw, d2phi/dw2 and d3phi/w3 can be calculated.
- For some materials also the extinction coefficient can be calculated.
- Unit tests to test accuracy of all formulas, extinction coefficients and spectral phase expansion terms.

PR #474, opened 2026-01-08
Fix for PR #369 refractive index

em-archer:
Modified setup.py to install .json files (e.g. refractive_index_db.json).

Update to PR #369 to automatically install database information stored in refractive_index_db.json. Small modification to setup.py so that the refractive data is automatically installed when using pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,11 @@ with open("./tests/requirements.txt") as f:
 setup(
     name="lasy",
     version=lasy.__version__,
-    packages=find_packages("."),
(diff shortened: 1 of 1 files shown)


rob-shalloo:
Hi Emily, thanks for finding this bug and for this PR. It looks good to me. I agree that in addition to the code throwing an error without the .json file being present,  the minimal set of materials (included in the .json file) should indeed be shipped by default.

rob-shalloo (approved):

PR #476, opened 2026-01-19
Add environment variable to switch backend

AlexanderSinn:
This PR adds the ability to switch the LASY backend using the environment variable `LASY_BACKEND`.
Available options are:
- `NP` (NumPy)
- `CP` (CuPy)
- `AUTO` (default) to try to import CuPy, but use NumPy if that fails.

A print statement is added announcing the chosen backend. Additionally, a `xp.is_available()` check is added to the try-except block to test if a GPU is installed. If not, a CUDARuntimeError is raised and NumPy is used. Previously, LASY would have tried to use CuPy and crashed in that case.

PR #477, opened 2026-01-29
[WIP] make from data profile GPU compatible

rob-shalloo:
The `LongitudinalProfileFromData` class take arrays as input which could be numpy arrays even when working on GPU. To account for this, we add the `to_gpu` function to the input data.

### DIFF
--- a/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
+++ b/lasy/profiles/longitudinal/longitudinal_profile_from_data.py
@@ -1,6 +1,6 @@
 from scipy.constants import c
 
-from lasy.backend import xp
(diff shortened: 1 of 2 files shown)


rob-shalloo:
Ultimately, would be better to link this to #476 and decide to cast data to GPU or CPU depending on the chosen backend which is more robust than `use_cuda` which will default to GPU even when operating on CPU (if `cupy` is installed)

AlexanderSinn:
With #476, `to_gpu` along with everything else exposed by `lasy.backend` will match the chosen backend. So I think this PR is fine as is.

MaxThevenet (approved):
LGTM too, @AlexanderSinn then we can just merge?

PR #478, opened 2026-02-03
Fixes for writing CW pulses to openPMD

em-archer:
Previously, writing the laser pulse to an openPMD file would fail in the case of a CW pulse. 

The laser pulse may still be saved as a field, but the user is warned that they are saving a CW pulse (since most PIC codes need temporal information). A CW pulse cannot be saved as a vector potential, so this is now properly caught.

### DIFF
--- a/lasy/laser.py
+++ b/lasy/laser.py
@@ -300,6 +300,8 @@ class Laser:
             self.grid,
             self.profile.lambda0,
             self.profile.pol,
+            self.profile.is_cw,
+            self.profile.is_plane_wave,
             save_as_vector_potential,
         )
         self.output_iteration += 1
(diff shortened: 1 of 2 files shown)


rob-shalloo:
Hey Emily, thanks for this! It's certainly good to be able to (and makes sense) that we can save a CW pulse to an openPMD file. Though, I would maybe suggest a slightly different approach. Rather than using a `try` `except` statement, could we instead try to probe the laser grid and check for the `is_cw` flag? Additionally, we could add the`is_plane_wave` flag. Then if both of those (and the general fail) then we could throw an error.

em-archer:
Now passes the attributes profile.is_cw and profile.is_plane wave to save non-3D laser data correctly. Tested for CW and plane wave lasers.

MaxThevenet (approved):
Looks good, thanks!

PR #479, opened 2026-02-03
Fix zernike RmnGenerator on GPU

AlexanderSinn:
#458 broke the `test_gsalgo.py` test when running the tests on GPU. For the `k` array we can simply use a numpy array as it should be small and not needed on the GPU. With a cupy array `math.factorial(n - i)` dosen't work because `i` is still a value on the GPU.

### DIFF
--- a/lasy/utils/zernike.py
+++ b/lasy/utils/zernike.py
@@ -1,4 +1,5 @@
 import math
+
 import numpy as np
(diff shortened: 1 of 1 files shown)


em-archer (approved):
Doing similar thing to closed PR #481

PR #482, opened 2026-04-09
Change default for vector potential to field conversion

MaxThevenet:
We noticed that the `direct` method to convert from vector potential to field, although it should be slightly more accurate physically, can be quite wrong if the longitudinal resolution isn't long enough. This PR changes the default to use the other method, which seems more robust.

### DIFF
--- a/lasy/utils/laser_utils.py
+++ b/lasy/utils/laser_utils.py
@@ -663,7 +663,7 @@ def field_to_vector_potential(grid, omega0):
     return -1j * e * grid.get_temporal_field() / (m_e * omega * c)
 
 
-def vector_potential_to_field(grid, omega0, direct=True):
+def vector_potential_to_field(grid, omega0, direct=False):
     """
     Convert envelope from electric field (V/m) to normalized vector potential.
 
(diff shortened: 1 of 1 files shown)


lasy/utils/laser_utils.py
```diff
@@ -663,7 +663,7 @@ def field_to_vector_potential(grid, omega0):
     return -1j * e * grid.get_temporal_field() / (m_e * omega * c)
 
 
-def vector_potential_to_field(grid, omega0, direct=True):
+def vector_potential_to_field(grid, omega0, direct=False):
     """
     Convert envelope from electric field (V/m) to normalized vector potential.
```
delaossa: The description of the method is wrong. It should be something like: "Convert envelope from normalized potential to electric field (V/m)."

delaossa (approved):
We tested the new default to retrieve the energy of the laser from an LPA simulation with Wake-T. It looks good! 
Thanks @MaxThevenet for spotting the issue.

lasy/utils/laser_utils.py
```suggestion
    Convert envelope from normalized vector potential to electric field (V/m).
```

AlexanderSinn (approved):

PR #485, opened 2026-04-27
Follow up on #469 to fix return of interpolated ```FromArrayProfile```

TimoEichner:
The issue observed by @AlexanderSinn in https://github.com/LASY-org/lasy/pull/469#issuecomment-4325607120 seems to be caused by an incorrect return of the `xyt` field after the interpolation method was adapted in #469. This PR adds a separate ```return combined_field```after evaluating the interpolated `xyt` field.


AlexanderSinn (approved):
I tested this locally, and this indeed fixes the issue with the hipace CI test!

PR #489, opened 2026-07-24
Interpolate real and imag part separately in FromArrayProfile

### LINKED ISSUE #488: Let's check that
https://github.com/LASY-org/lasy/blob/41c653e5fcea814daee72f20746aaf00d0c59f94/lasy/profiles/from_array_profile.py#L114

AlexanderSinn:
When doing RZ -> 3D or RZ -> RZ at a high resolution, interpolating abs and phase causes artifacts.

### DIFF
--- a/lasy/profiles/from_array_profile.py
+++ b/lasy/profiles/from_array_profile.py
@@ -94,8 +94,7 @@ class FromArrayProfile(Profile):
                 self.field_interp_modes.append(
                     RegularGridInterpolator(
                         (r, axes["t"]),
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
Doing more Axicon simulations with Wake-T (which uses the interpolation from LASY), the current dev version also gives bad results, and this PR fixes it.

MaxThevenet (approved):
Thanks for this PR! Let's give this a try as is, and if something comes up we'll support both options.

PR #490, opened 2026-08-28
Fresnel transform chirped-z propagator for RT geometry

em-archer:
Chirped-z propagator that uses a Hankel transform whose sampling is dictated by the desired output grids, in order to perform the zoom analogously to scipy's zoom fft. 

Currently written as a separate file and module, we may wish to consider combining this with the existing Fresnel chirped-z with a flag to specify geometry. Can also make everything run on GPU, I just didn't implement this yet but instead bring back to CPU for one operation.


em-archer:
Merged into one propagator that works for either geometry and now also working entirely on backend by importing j0 (Bessel function) from lasy.backend

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -121,8 +166,6 @@ def update(self, dim, omega0):
         self.dim = dim
         self.omega0 = omega0
 
-        assert dim in ["xyt"], "Invalid dimension. Only 'xyt' is currently supported."
```
MaxThevenet, line 124: Could we keep it? e.g. 
```
        assert dim in ['xyt', 'rt'], "Invalid dimension, must be 'xyt' or 'rt'"
```

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -164,11 +204,57 @@ def _zoomFourierTransform2D(self, x, y, f, k_x, k_y):
...
+
+        Setting ``k_r = k * r_out / z`` evaluates the transform at exactly
+        the spatial frequencies corresponding to the output grid positions,
+        mirroring how the Chirp-Z transform evaluates the DFT at a freely
+        chosen set of frequencies rather than the standard FFT grid.
+
+        Parameters
+        ----------
+        r : array_like, shape (N,)
+            Uniformly-spaced radial coordinates of the input field
```
MaxThevenet, line 235:
```suggestion
            Uniformly spaced radial coordinates of the input field
```

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -164,11 +204,57 @@ def _zoomFourierTransform2D(self, x, y, f, k_x, k_y):
...
+            (must start at or very near zero).
+
+        f : array_like, shape (N,)
+            Complex field values at the radial positions ``r``.
+
+        k_r : array_like, shape (M,)
+            Radial spatial frequencies (rad m⁻¹) at which to evaluate the
+            transform.  Typically ``k_r = k * r_out / z``.
```
MaxThevenet, line 243:
```suggestion
            transform. Typically ``k_r = k * r_out / z``.
```

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -197,69 +283,90 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
         self.update(dim, omega0)
 
+        # --- Common setup ---
         initial_position = grid_in.position
-
-        # Get the spectral field from the grid objects
         field_in, omega = grid_in.get_spectral_field()
+
         if grid_out is None:
-            # Create a new grid for the output if not provided
```
MaxThevenet, line 205: Is there a reason why these comments don't apply anymore? Otherwise, I think we can keep them.

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -197,69 +283,90 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
             grid_out = copy.deepcopy(grid_in)
             grid_out.set_spectral_field(xp.zeros_like(field_in))
         field_out = grid_out.spectral_field
-        omega += omega0
+
+        omega = omega + omega0  # avoid mutating the grid's internal array
```
MaxThevenet, line 295: Isn't it the same? If so, I'd recommend we keep as is if the change is unrelated to the new rt implementation. Also, the function `get_spectral_field` returns a copy https://github.com/LASY-org/lasy/blob/55f7b55946a5f9f6876b47076025400cf9ed71fb/lasy/utils/grid.py#L235.

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -197,69 +283,90 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
-            )
+                field_out[:, :, indx] = F * postFactor
+
+        elif self.dim == "rt":
+            r = grid_in.axes[0]
+            rF = grid_out.axes[0]
+
+            for indx in indxs:
+                om = omega[indx]
+                k = om / c
```
MaxThevenet, line 346:
```suggestion
                k = omega[indx] / c
```
Unless I missed something, I don't think `om` is re-used below (also, the name `om` isn't very specific).

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -197,69 +283,90 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
+                k_r = k * rF / distance
+
+                # field_in has shape (Nr, n_azimuthal_modes, Nω); squeeze out the
+                # modes axis (=1) so _zoomHankelTransform receives a 1-D vector.
+                F = self._zoomHankelTransform(
+                    r, xp.squeeze(field_in[:, :, indx]) * prefactor, k_r
+                )
+
+                postFactor = (
+                    (-1j * k / distance)
```
MaxThevenet, line 358: Parentheses can be removed (or not, up to you).

lasy/propagators/fresnel_chirpztransform_propagator.py
```diff
@@ -197,69 +283,90 @@ def propagate(self, grid_in, dim=None, omega0=None, distance=None, grid_out=None
...
-            # Add output field to array
-            field_out[:, :, indx] = F * postFactor
+                field_out[:, :, indx] = F * postFactor
 
-        # Shift the pulse back to the center of the time axis
-        field_out *= xp.exp(-1j * omega[xp.newaxis, xp.newaxis, :] * distance / c)
+        # --- Common teardown ---
+        # Shift pulse back to centre of time axis; broadcast omega over all spatial axes
+        omega_bc = omega.reshape((1,) * (field_out.ndim - 1) + (-1,))
+        field_out *= xp.exp(-1j * omega_bc * distance / c)
```
MaxThevenet, line 368: Is this doing the same operation than before for the xyt geometry?

MaxThevenet (changes requested):

lasy/propagators/fresnel_chirpztransform_propagator.py
em-archer: Essentially yes, because we always have two spatial dimensions (this would be more robust in cases where you don't).

MaxThevenet (approved):
LGTM, thanks!

PR #492, opened 2026-09-02
When doing laser.show(), move the axis from gpu to cpu for matplotlib plotting

i-linyeh:


MaxThevenet:
Thanks for the PR! Docs build fails https://app.readthedocs.org/projects/lasydoc/builds/34488374/ could you have a look?

PR #494, opened 2026-09-09
Fix bug reading laser from HIPACE++

huixingjian:
```FromOpenPMDProfile``` currently raises ```ErrorNoSuchAttribute``` when the input file lacks the angularFrequency attribute, preventing users from loading the laser envelope even when its central frequency is known.
This PR adds an optional omega0 argument to provide the central angular frequency in rad/s when this metadata is missing.
Raises a clear ValueError if neither is available, guiding users to provide omega0 manually.
Rejects non-positive fallback frequencies and documents the new argument.

### DIFF
--- a/lasy/profiles/from_openpmd_profile.py
+++ b/lasy/profiles/from_openpmd_profile.py
@@ -56,10 +56,28 @@ class FromOpenPMDProfile(FromArrayProfile):
             m = it.meshes[envelope_name]
             geometry = m.get_attribute("geometry")
             dim = "xyt" if geometry == "cartesian" else "rt"
-            if omega0 == None :
-                omg0 = m.get_attribute("angularFrequency")
-            else:
-                omg0 = omega0
+            try:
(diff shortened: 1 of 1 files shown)


MaxThevenet:
If the input file does not have the angular frequency metadata, it does not comply with the openpmd standard. I would argue, the file is then incomplete, because an envelope without its reference frequency is ill-defined. How was this file generated?

huixingjian:
> If the input file does not have the angular frequency metadata, it does not comply with the openpmd standard. I would argue, the file is then incomplete, because an envelope without its reference frequency is ill-defined. How was this file generated?

It's actually laser output from HIPACE++

MaxThevenet:
Then there's an issue that we need to figure out. HiPACE++ should write this attribute at https://github.com/Hi-PACE/hipace/blob/92d3da222f62481e9eacf3207b98aa22fea47b62/src/diagnostics/OpenPMDWriter.cpp#L235.

